PR #1685, opened 2017-01-13
Notary optimisation : add start service button from the clients table

kamsi:
Following #1682, this is point number 4 of the notary optimisation.

In the clients table, have a button "Start a service for this client" for each client. That button will be just between the quantities of services started by client and the "Access to the clients account" button. 

minegocio_gt

On click on this button, a modale window would open with a choice of services that the notary can run for the client.

This is the less-important issue of notary-optimisation milestone, do it only if easy.

### DIFF
--- a/controller/manager/index.js
+++ b/controller/manager/index.js
@@ -61,6 +61,37 @@ exports['clients/[0-9][a-z0-9]+/delete'] = {
 	match: managedUserMatcher
 };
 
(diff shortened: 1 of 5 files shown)


kamsi:
@medikoo let's wait with merge until adjusts are approved: https://github.com/egovernment/eregistrations-salvador/pull/1677 , https://github.com/egovernment/eregistrations-guatemala/pull/1013

mtuchowski:
Both adapts reviewed.

PR #1750, opened 2017-01-17
Enable default properties lists in eregistrations

kamsi:
Its very inconvinient having to rely on properties lists in the end systems. Hence, we should move thelists to eregistrations and only enable their mutation in the end systems. Its about following lists:

- `apps-common/business-process-list-properties.js`
- `apps-common/business-process-list-computed-properties.js`
- `apps-common/user-list-properties.js`
- `apps-common/business-process-dispatcher-list-properties.js`


kamsi:
@medikoo adjusts in review

kamsi:
@medikoo I will pass adjusts to review after approval

medikoo:
Looks good, we may either merge it now, or merge it after testing it with adapts, let me know

kamsi:
@medikoo let's wait for adapts

mtuchowski:
@medikoo Concerning `businessName` property. By default it is a static property (https://github.com/egovernment/eregistrations/blob/master/model/business-process-new/base.js#L43) but in the end systems the usage vary:

* Salvador - only computed.
* Guatemala - both computed and static, depending on service.
* Tanzania - both computed and static, depending on service.

Which list in eregistrations should contain this property? We talked with Kamil and think it should be in both by default to minimize work and bugs in end-systems - in worst case scenario (Salvador) it will be pointless for static list.

medikoo:
@mtuchowski In eregistrations its value is empty as its value is service specific.
Occasionally (but I think rarely) it happens that we use this field directly in data forms and it such case it remains static. However in most cases it's linked to some other meaningful property as e.g. `companyName`.

I'd say that easiest and least error-prone approach would be to unconditionally index it  as computed record for any service , and always have it in computed-list. It will work also if property handling is purely static.

mtuchowski:
All adapts OK.

PR #1751, opened 2017-01-18
Introduce officialFullName to status logs.

mtuchowski:
Requested in https://github.com/egovernment/eregistrations-guatemala/issues/810.

### DIFF
--- a/model/requirement-upload.js
+++ b/model/requirement-upload.js
@@ -155,7 +155,6 @@ module.exports = memoize(function (db) {
 				statusLog.push({
 					label: log.getOwnDescriptor('label').valueToJSON(),
 					time: log.getOwnDescriptor('time').valueToJSON(),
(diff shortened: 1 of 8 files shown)


model/requirement-upload.js
```diff
@@ -155,6 +155,7 @@ module.exports = memoize(function (db) {
 				statusLog.push({
 					label: log.getOwnDescriptor('label').valueToJSON(),
 					time: log.getOwnDescriptor('time').valueToJSON(),
+					officialFullName: log.getOwnDescriptor('officialFullName').valueToJSON(),
```
medikoo: We do not store `officialFullName` in status logs for uploads, it's only certificates that have them at this point, still as I see we do not include statusLog for certs snapshot (that's probably bug, I've opened issue for that at #1754)

PR #1754, opened 2017-01-19
Include certificate status log in snapshot

kamsi:
I think we forgot about that, when we added status logs to certificates at #878 

Then they also should be read from snapshot when we visit certificate that has snapshot

### DIFF
--- a/model/document.js
+++ b/model/document.js
@@ -133,11 +133,8 @@ module.exports = memoize(function (db) {
 				label: this.database.resolveTemplate(this.label, this.getTranslations(), { partial: true }),
 				abbr: this.abbr,
 				status: this.status,
-				overviewSection: this.overviewSection.toJSON(),
-				statusLog: this.statusLog.ordered.map(function (stautLogEntry) {
-					return stautLogEntry.toJSON();
-				}).toArray()
-			};
+				overviewSection: this.overviewSection.toJSON()
+			}, statusLog = [];
 			if (this.issuedBy) data.issuedBy = this.getOwnDescriptor('issuedBy').valueToJSON();
 			if (this.issueDate) data.issueDate = this.getOwnDescriptor('issueDate').valueToJSON();
 			if (this.number) data.number = this.getOwnDescriptor('number').valueToJSON();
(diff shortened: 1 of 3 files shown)


kamsi:
@medikoo views are already prepared to handle such snapshot (tested).

model/document.js
```diff
@@ -133,7 +133,10 @@ module.exports = memoize(function (db) {
 				label: this.database.resolveTemplate(this.label, this.getTranslations(), { partial: true }),
 				abbr: this.abbr,
 				status: this.status,
-				overviewSection: this.overviewSection.toJSON()
+				overviewSection: this.overviewSection.toJSON(),
+				statusLog: this.statusLog.ordered.map(function (stautLogEntry) {
```
medikoo: Typo

model/lib/status-log.js
```diff
@@ -13,6 +13,15 @@ module.exports = memoize(function (db) {
 		time: { type: db.DateTime, required: true },
 		official: { type: User },
 		officialFullName: { type: StringLine },
-		text: { type: db.String, required: true }
+		text: { type: db.String, required: true },
+		toJSON: { value: function (ignore) {
+			return {
+				label: this.getOwnDescriptor('label').valueToJSON(),
+				time: this.getOwnDescriptor('time').valueToJSON(),
+				official: this.getOwnDescriptor('official').valueToJSON(),
```
medikoo: No need for `official`

model/lib/status-log.js
```diff
@@ -13,6 +13,15 @@ module.exports = memoize(function (db) {
...
+				officialFullName: this.getOwnDescriptor('officialFullName').valueToJSON(),
```
medikoo: We should add it only if we have `officialFullName` (e.g. it doesn't apply to requirement uploads now).
Otherwise we create records as `officialFullName: undefined` which bring noise to snapshot and unjustifiably increase their size.

model/document.js
medikoo: Also this should not be retrieved by creating an reactive array instance, but more as here -> https://github.com/egovernment/eregistrations/blob/master/model/requirement-upload.js#L153-L161

model/lib/status-log.js
```diff
@@ -13,6 +13,15 @@ module.exports = memoize(function (db) {
...
 		official: { type: User },
 		officialFullName: { type: StringLine },
-		text: { type: db.String, required: true }
+		text: { type: db.String, required: true },
+		toJSON: { value: function (ignore) {
+			return {
+				label: this.getOwnDescriptor('label').valueToJSON(),
+				time: this.getOwnDescriptor('time').valueToJSON(),
+				officialFullName: this.officialFullName
+					&& this.getOwnDescriptor('officialFullName').valueToJSON(),
```
medikoo: This doesn't change anything, still we end with `officialFullName: undefined` which we don't want

PR #1760, opened 2017-01-24
Allow inserting suctom sections in official payment views.

mtuchowski:
Just like for requirement upload custom sections, we need the possibility to append some sections to business-process-official-payment and business-process-revision-payment views.


medikoo:
There are some lint issues

PR #1757, opened 2017-01-25
Stats: Tables: `statusHistory` introduction

kamsi:
Needed for:

https://docs.google.com/document/d/1i92UhTr_krhLYVc9TA88WGIrvwPAnM9x84WHCFpVtQc/edit

In order to be able to get all the information (like how many pendings were there on a given day) for the new stats we need to gather some more data. I propose to create `ProcessingStep.prototype.flowLog` as a nested map. It's entries must contain the following:

`status` - Status that was setup for the step.
`stamp` - Moment when given status change was triggered.

The setup should be performed from within the current flow service: https://github.com/egovernment/eregistrations/blob/master/server/services/business-process-flow.js

All the status changes should be written to the `flowLog` of corresponding steps.

### DIFF
--- a/model/business-process-new/flow.js
+++ b/model/business-process-new/flow.js
@@ -109,26 +109,22 @@ module.exports = memoize(function (db/*, options*/) {
 		} },
 		// Computed status of registration
 		status: { type: BusinessProcessStatus, value: function (_observe) {
-			var frontDesk, result;
+			var frontDesk;
 			if (!this.isSubmitted) return 'draft';
 			if (this.isSentBack) return 'sentBack';
 			if (this.isRejected) return 'rejected';
 			if (this.isClosed) return 'closed';
 
-			_observe(this.processingSteps.applicable).some(function (step) {
-				if (_observe(step._status) === 'rejected') {
-					result = step.status;
-					return true;
-				}
-			});
-			if (result) return result;
-			this.processingSteps.applicable.some(function (step) {
-				if (_observe(step._status) === 'sentBack' && _observe(step._isPreviousStepsSatisfied)) {
-					result = step.status;
-					return true;
-				}
-			});
-			if (result) return result;
+			if (_observe(this.processingSteps.applicable).some(function (step) {
+					return _observe(step._status) === 'rejected';
+				})) {
+				return 'rejected';
+			}
+			if (this.processingSteps.applicable.some(function (step) {
+					return _observe(step._status) === 'sentBack' && _observe(step._isPreviousStepsSatisfied);
+				})) {
+				return 'sentBack';
+			}
 			if (this.processingSteps.applicable.every(function (step) {
 					return _observe(step._status) === 'approved';
 				})) {
--- a/model/business-process-new/flow.js
+++ b/model/business-process-new/flow.js
@@ -131,6 +131,10 @@ module.exports = memoize(function (db/*, options*/) {
 				return 'closed';
 			}
 
+			frontDesk = this.processingSteps.frontDesk;
+			if (frontDesk && _observe(frontDesk._isApproved)) return 'withdrawn';
+			if (frontDesk && _observe(frontDesk._isPending)) return 'pickup';
+
 			if (this.processingSteps.revisions.some(function (processingStep) {
 					return _observe(processingStep._isPending) ||
 						(_observe(processingStep._status) === 'approved' &&
@@ -138,10 +142,6 @@ module.exports = memoize(function (db/*, options*/) {
 				})) {
 				return 'revision';
 			}
-
-			frontDesk = this.processingSteps.frontDesk;
-			if (frontDesk && _observe(frontDesk._isApproved)) return 'withdrawn';
-			if (frontDesk && _observe(frontDesk._isPending)) return 'pickup';
 
 			return 'process';
 		} },
--- a/model/business-process-new/flow.js
+++ b/model/business-process-new/flow.js
@@ -115,20 +115,26 @@ module.exports = memoize(function (db/*, options*/) {
 			if (this.isRejected) return 'rejected';
 			if (this.isClosed) return 'closed';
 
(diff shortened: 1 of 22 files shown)


kamsi:
@medikoo let's have this approved before I test it in all systems, and prepare adjusts (only thing needed should be initialization in master process).

model/lib/status-history-log.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+var memoize          = require('memoizee/plain')
+  , defineStringLine = require('dbjs-ext/string/string-line');
+
+module.exports = memoize(function (db) {
+	var StringLine = defineStringLine(db);
+
+	return db.Object.extend('StatusHistoryLog', {
```
medikoo: Maybe let's name it `StatusHistoryItem`, it should not be confused with `statusLog` which is similar but different

server/services/status-history-logger.js
```diff
@@ -0,0 +1,60 @@
...
+  , Set                           = require('es6-set')
+  , processingStepsMeta           = require('../../processing-steps-meta')
+  , uuid                          = require('time-uuid')
+  , uniqIdPrefix                  = 'abcdefghiklmnopqrstuvxyz'[Math.floor(Math.random() * 24)];
+
+var storeLog = function (storage, logPath) {
+	return function (event) {
+		var status, oldStatus;
+
+		if (event.type !== 'computed' || !event.old) return;
```
medikoo: We shouldn't discard record if it's first record in history for this property

test/model/lib/status-history-log.js
```diff
@@ -0,0 +1,3 @@
+// TODO
+
+'use strict';
```
medikoo: Let's add some sanity test

medikoo:
@kamsi conflicts to be resolved

server/services/status-history-logger.js
```diff
@@ -0,0 +1,72 @@
...
+
+var storeLog = function (storage, logPath/*, options */) {
+	var options = Object(arguments[2]);
+	return function (event) {
+		var status, oldStatus, resolvedPath, logPathResolved;
+		logPathResolved = logPath ? logPath + '/' + getPathSuffix() : getPathSuffix();
+
+		if (event.type !== 'computed') return;
+		status    = unserializeValue(event.data.value) || null;
+		oldStatus = (event.old && unserializeValue(event.old.value)) || null;
```
medikoo: As we discussed let's not store `null` but `''` (`null` is not acceptable value for `cardinalPropertyKey`)

server/services/status-history-logger.js
```diff
@@ -0,0 +1,72 @@
...
+
+		// We ignore such cases, they may happen when direct overwrites computed
+		if (status === oldStatus) return;
+		if (options.processorPath) {
+			storage.get(event.ownerId + '/' + options.processorPath)(function (data) {
+				if (data && data.value[0] === '7') {
+					resolvedPath = event.ownerId + '/' + logPathResolved + '/processor';
+					storage.store(resolvedPath, data.value);
+					debug('Stored processor %s for the path %s', data.value, resolvedPath);
```
medikoo: No need for those logs, as db logs, show all stores

server/services/status-history-logger.js
```diff
@@ -0,0 +1,72 @@
...
+				}
+			});
+		}
+		resolvedPath = event.ownerId + '/' + logPathResolved + '/status';
+		storage.store(resolvedPath, serializeValue(status));
+		debug('Stored status %s for the path %s', status, resolvedPath);
```
medikoo: Same here

server/services/status-history-logger.js
```diff
@@ -0,0 +1,72 @@
...
+		status    = unserializeValue(event.data.value) || null;
+		oldStatus = (event.old && unserializeValue(event.old.value)) || null;
+
+		// We ignore such cases, they may happen when direct overwrites computed
+		if (status === oldStatus) return;
+		if (options.processorPath) {
+			storage.get(event.ownerId + '/' + options.processorPath)(function (data) {
+				if (data && data.value[0] === '7') {
+					resolvedPath = event.ownerId + '/' + logPathResolved + '/processor';
+					storage.store(resolvedPath, data.value);
```
medikoo: This returns promise which should be handled

server/services/status-history-logger.js
```diff
@@ -0,0 +1,72 @@
...
+					debug('Stored processor %s for the path %s', data.value, resolvedPath);
+				}
+			});
```
medikoo: Returns promise that should be handled

server/services/status-history-logger.js
```diff
@@ -0,0 +1,72 @@
...
+		}
+		resolvedPath = event.ownerId + '/' + logPathResolved + '/status';
+		storage.store(resolvedPath, serializeValue(status));
```
medikoo: Returns promise that should be handled

model/business-process-new/flow.js
```diff
@@ -109,12 +109,32 @@ module.exports = memoize(function (db/*, options*/) {
...
 			if (this.isRejected) return 'rejected';
 			if (this.isClosed) return 'closed';
 
+			_observe(this.processingSteps.applicable).some(function (step) {
+				if (_observe(step._status) === 'rejected') {
+					result = step.status;
+					return true;
+				}
+			});
+			if (result) return result;
```
medikoo: Looks a bit overcomplicated. It seams that cleaner would be:

```javascript
if (_observe(this.processingSteps.applicable).some(function (step) {
    return _observe(step._status) === 'rejected';
  })) {
  return 'rejected';
}
```

test/model/lib/status-history-item.js
```diff
@@ -0,0 +1,15 @@
+// TODO
```
medikoo: Let's remove that comment

Issue #1742, opened 2017-01-11
Reject modal redesign


mtuchowski:
@vianneyl We had to revert https://github.com/egovernment/eregistrations/pull/1461 because it made the reject dialog not working.

@melux Could you try to redo https://github.com/egovernment/eregistrations/pull/1461 making sure that the input stays inside the form https://github.com/egovernment/eregistrations/pull/1461/files#diff-5a4694dc54c33d6f87e101768987d7dcR89

PR #1740, opened 2017-01-10
Rejection confirmation for officials.

mtuchowski:
As described in https://github.com/egovernment/eregistrations-salvador/issues/1483.

PR #1762, opened 2017-01-26
Problems with rejection dialog.

mtuchowski:
Again, after https://github.com/egovernment/eregistrations/pull/1461 was reverted, as explained in https://github.com/egovernment/eregistrations/issues/1742 and https://github.com/egovernment/eregistrations/pull/1740 was meant to fix it, the rejection dialog for revision official is broken.


mtuchowski:
@kamsi This ticket will fix the problem you reported earlier today.

PR #1773, opened 2017-01-30
Allow custom names for `frontDesk` step

kamsi:
While we do not envision to have more than one front desk step per service, it will make configuration of services easier if we could have frontDesk on one service named differently than on other.

Only through such configuration we can out of a box introduce separation of two steps as those that should be handled by two different entities. Trying to achieve it using same step name, forces cumbersome customizations.

What should be done:
- Introduce `processingSteps.frontDesk` property which by default should resolve to `processingSteps.map.frontDesk`
- In calculation of `businessProcess.status` rely on `processingSteps.frontDesk` and not on `processingSteps.map.frontDesk`
- In [certificates-status-log](https://github.com/egovernment/eregistrations/blob/master/server/services/certificates-status-log.js#L97-L98) service ensure to resolve `whithdrawalConfig` per `BusinessProcess` type, and on basis of its `processingSteps.frontDesk` resolve `triggerPath` an `offficialPath`
- In [pickup-institution-form-section](https://github.com/egovernment/eregistrations/blob/master/view/dbjs/pickup-institution-form-section.js#L12) DOM binding ensure to rely on `processingSteps.frontDesk`
- In [statistics-files-pending](https://github.com/egovernment/eregistrations/blob/master/view/statistics-files-pending.js#L102-L107) Update resolution of frontDesk
- In [statistics-dashboard](https://github.com/egovernment/eregistrations/blob/master/view/statistics-dashboard.js#L235-L332) Update resolution of frontDesk

Having that we should test and check eventual customizations in end systems.


kamsi:
We need adjsuts.

view/statistics-files-pending.js
```diff
@@ -102,8 +103,20 @@ exports['statistics-main'] = function () {
...
-							return data ? data._pending : null;
+							var result = null;
+							if (!getFrontDeskNames().size) return null;
+							getFrontDeskNames().forEach(function (name) {
+								var value = data.atPartB.getBySKeyPath(resolveFullStepPath(name));
+								if (value) {
+									if (!result) {
+										result = value._pending;
+									} else {
+										result.value += value.pending;
```
medikoo: I think you modify here object which should not be modified, you take object that resolves value for single step, and try to add there value from other step. It should technically crash, if it doesn't then it's even worse (data gets written to server?)

view/statistics-files-pending.js
```diff
@@ -102,8 +103,20 @@ exports['statistics-main'] = function () {
 			this.processingStepsMeta.frontDesk ?
 					generateRow({ class: 'statistics-table-sub-header statistics-table-sub-header-success' },
 						_("Pending for withdraw at Front Desk"), services, function (data) {
-							data = data.atPartB.getBySKeyPath(resolveFullStepPath('frontDesk'));
-							return data ? data._pending : null;
+							var result = null;
+							if (!getFrontDeskNames().size) return null;
```
medikoo: We should not have that check, instead we should fix condition at 103 (` 		this.processingStepsMeta.frontDesk ?`)

view/statistics-files-pending.js
```diff
@@ -99,11 +100,22 @@ exports['statistics-main'] = function () {
 				function (data) { return data._rejected; }, this.processingStepsMeta,
 				"statistics-table-sub-header-sentback", function (data) { return data.rejected; }),
 
-			this.processingStepsMeta.frontDesk ?
+			getFrontDeskNames().size ?
```
medikoo: It would be cleaner to save the result of `getFrontDeskNames()` and reuse it in those two places

kamsi:
@medikoo yes I can. I haven't noticed anything suspicious.

mtuchowski:
Adjust looks OK.

PR #1709, opened 2017-01-30
Superuser and messages

mtuchowski:
Hello team, 

I add an other particularity of `superuser` accounts here. If easy it would be great to set the notifications so that the superuser does not receive any emails. Let me know!

### DIFF
--- a/server/services/query-memory-db/queries/notify-business-process-assigned.js
+++ b/server/services/query-memory-db/queries/notify-business-process-assigned.js
@@ -4,11 +4,12 @@ var ensureDatabase = require('dbjs/valid-dbjs')
   , notifyAssigned = require('../../../email-notifications/business-process-assigned');
 
 module.exports = function (db) {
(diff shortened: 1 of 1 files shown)


server/services/query-memory-db/queries/notify-business-process-assigned.js
```diff
@@ -4,8 +4,9 @@ var ensureDatabase = require('dbjs/valid-dbjs')
   , notifyAssigned = require('../../../email-notifications/business-process-assigned');
 
 module.exports = function (db) {
-	var dispatcherEmails = ensureDatabase(db).User.filterByKey('roles', function (roles) {
+	var dispatcherEmails = ensureDatabase(db).User.filterByKey('roles', function (roles, user) {
 		if (!roles) return false;
+		if (user.isSuperUser) return false;
```
medikoo: This won't work as change of `isSuperUser` will not update collection.

It has to be done possibly as union of two collections: `dispatchers.and(notSuperUsers)`

Issue #1759, opened 2017-01-24
Stats: Tables: View interfaces


kamsi:
https://github.com/egovernment/eregistrations/issues/1756

This is about preparing views and controllers for the new statistics: https://docs.google.com/document/d/1i92UhTr_krhLYVc9TA88WGIrvwPAnM9x84WHCFpVtQc/edit

The data will be handled separately here: https://github.com/egovernment/eregistrations/issues/1758

medikoo:
@vianneyl do views as presented here: http://xwv4ez.axshare.com/#p=table_of_quantities_of_files_in_status__per_certif
are final (?)

I think there were some changes to them agreed, it'll be great to have them updated.

vianneyl:
@medikoo yes they are final

2017-01-26 16:08 GMT+01:00 Mariusz Nowak <notifications@github.com>:

medikoo:
@vianneyl some clarifications

__All tables__

As in case of some other statistics pages, we will operate only on data gathered after some point in time. Therefore we should gently inform user about with it message as:
_As history of status changes is properly recorded since 1st of February 2017, below table only exposes data for files processed after that day_.
I take that we should follow what was proposed here: #1727 and put that info into (i) tooltip by period selector.

__Table 1 & Table 2__

It's not shown on view, but I take we will paginate results to not show more than 100 rows per page (paginator will only show if we have more than 100 items)

__Table 3__

- Here pagination is not straighforward, e.g. for first table we may have 10000 results and in other 100 results, it's not clear how to deal with pagination here.
I take that each table should be paginated individually, but then it becomes more challenging to reflect that in url address bar (which should be in place if we want to have that page working properly in all supported browsers). I propose that we should introduce query names as `revisionPage`, `precalPage` etc.
-  I understand that we show only those operators which have some processing history at given day. So there can be scenarios where not only some processors will appear from time to time, but some days may not appear at all.
- In case where either "Sent back.." or "Rejected" do not apply to given step we do not show this column (alternative way is to show "N/A" in them, but I think it's unnecessary noise)

@vianneyl please confirm is all that ok, or elaborate if you have other ideas

Additionally it's not clear how to prepare CSV export for Table 3, can you prepare some proposal?

medikoo:
@vianneyl  other question, in spec you write:

```
The lines of the table are: 
- days included in the period selector  if “Daily” is selected
- weeks included in the selected period if “weekly” is selected
- months included in the selected period if “Monthly” is selected
[...]

PR #1764, opened 2017-01-31
Stats: Tables: Certificates: HTML view

kamsi:
Specification: https://docs.google.com/document/d/1i92UhTr_krhLYVc9TA88WGIrvwPAnM9x84WHCFpVtQc/edit

Additional important details to be found at #1759

UX mockup in http://xwv4ez.axshare.com/#p=table_of_quantities_of_files_in_status__per_certif

It should come with server GET controller that will retrieve needed data using utilities configured at #1758 

The filter bar should expose on url query with: `service`, `certificate`, `dateFrom`, `dateTo` and `mode` (either _daily_, _weekly_, _monthly_ or _yearly_).

View is expected to query server with `dateFrom`, `dateTo`,  `mode` (either _daily_, _weekly_, _monthly_ or _yearly_) and `page` params. The _service_ and _certificate_ filter should be resolved locally (server will respond with data for all combinations).
Result of server queries should be cached on client, we should not query server with same question twice (there's no chance result will change).

The data format as returned by server controller to client was already agreed at #1758, it'll be as:

for daily `mode`:

```javascript
{
  "2016-01-02": periodSums,
  "2016-01-03": periodSums,
  "2016-01-04": periodSums
  etc.
}
```
or for monthly `mode`:
```javascript
{
  "2016-01": periodSums, 
  "2016-02": periodSums,
  "2016-03": periodSums
  etc.
}
```

Where `periodSums` should most likely be reduced results of `calcuateStatusEventsSums` (as implemented at #1758)

The _direct_ result of `calcuateStatusEventsSums` would be as:

```
data.businessProcess[name][status] = sum
data.certificate[name][status] = sum
data.processingStep[name][processorId][status] = sum
```

While for this view we need to have it reduced to:

```
data.businessProcess[name][status] = sum
data.certificate[name][status] = sum
```

The normalization should be done on server (so client receives it already normalized) as we will also need same data format for PDF and CSV views as generated on server.


server/routes/statistics.js
```diff
@@ -66,6 +66,12 @@ module.exports = function (config) {
 	getData(driver).done();
 
 	return assign({
+		'get-flow-data': function (query) {
+			return queryHandler.resolve(query)(function (query) {
```
medikoo: We should use different queryHandler for this request, query options for requests are different, and validation rules are different

In `queryHandler` as provided here, we miss proper normalization and validation of e.g. `page`. so it's count does not overflow the pages count (e.g. see one made in business process table query handler: https://github.com/egovernment/eregistrations/blob/master/view/components/business-processes-table/setup-query-handler.js#L61-L62)

Then we should use same `queryHandler` on both _client_ and _service_ (but again: trying to reuse one configured for quite different filter bar in different view, won't work well).

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,331 @@
...
+	var result;
+	switch (mode) {
+	case 'weekly':
+		result = Math.ceil(duration.days / 7);
+		break;
+	case 'monthly':
+		result = duration.months;
+		break;
+	case 'yearly':
+		result = duration.years;
```
medikoo: It seems not right for all apart `daily`.

On example: if _mode_ is `monthly` and duration is from `2016-03-29` to `2016-05-03`, then it should report `3` months (`2016-03`, `2016-04` and `2016-05`) while `duration.months` will return `1`.

Similarly with weekly we need to settle on fixed mapping that each week starts on Monday and ends on Sunday

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,331 @@
...
+
+	return result;
+};
+
+exports['statistics-main'] = function () {
+	var queryHandler, data = new ObservableValue({}), pagination = new Pagination('/flow/');
+	queryHandler = setupQueryHandler(getQueryHandlerConf({
+		processingStepsMeta: this.processingStepsMeta
+	}), location, '/flow/');
+	queryHandler.on('query', function (query) {
```
medikoo: We should use dedicated query handler for this filter, reusing other one configured for other filter bars won't work well. The code here ideally should be as simple as:

```javascript
queryHandler.on('query', function (query) {
  // Query handler's job is to provide us with validated and normalized query object.
  // Ideally we don't have to do any further adaptation, but we have small exceptions as:
  ... convert date instances back to date strings ...
  // then we can just provide query to server, so:
  queryServer(serverQuery).done(...);
}

File: view/statistics-flow-certificates.js [hunk 4]
```diff
@@ -0,0 +1,331 @@
...
+						radios = container.getElementsByTagName('input');
+						container.addEvent('change', function () {
+							setTimeout(onChange, 0);
+						});
+						container.addEvent('click', function () {
+							setTimeout(onChange, 0);
+						});
+						onChange();
+					}, 0);
+				}("mode-selection")))),
```
medikoo: What is this `script` doing here? No chance to reuse (or refactor for reuse) what we have made for button group?

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,331 @@
...
+
+	section({ class: 'section-primary users-table-filter-bar' },
+		form({ action: '/flow/', autoSubmit: true },
+			div({ class: 'users-table-filter-bar-status' },
+				select(
+					{ id: 'service-select', name: 'service' },
+					option({ value: 'all', selected: location.query.get('service').map(function (value) {
+						return !value || value === 'all' ? 'selected' : null;
+					}) }, _("All services")),
+					list(db.BusinessProcess.extensions, function (ServiceType) {
```
medikoo: It would be good to create is as reusable component (it's not only for this view, and I think we already have views that have _service_ filter)

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,331 @@
...
+			data.value = filterData(responseData);
+		});
+	});
+
+	section({ class: 'section-primary users-table-filter-bar' },
+		form({ action: '/flow/', autoSubmit: true },
+			div({ class: 'users-table-filter-bar-status' },
+				select(
+					{ id: 'service-select', name: 'service' },
+					option({ value: 'all', selected: location.query.get('service').map(function (value) {
```
medikoo: `value: ''` (it's what we use for default selection)

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,331 @@
...
+								var selected = (serviceName ? (value === serviceName) : (value == null));
+								return selected ? 'selected' : null;
+							})
+						}, ServiceType.prototype.label);
+					})
+				)),
+			div({ class: 'users-table-filter-bar-status' },
+				select(
+					{ id: 'certificate-select', name: 'certificate' },
+					option({ value: '', selected: location.query.get('certificate').map(function (value) {
```
medikoo: Same here, some reusable component will be good (e.g. in inpector view we should have also _certificates_ filter, even though now there is _registrations_ filter)

view/statistics-flow-certificates.js
kamsi: Well, not exactly so:

image

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,303 @@
...
+		var serverQuery = copy(query), dateFrom, dateTo, mode
+		  , currentDate, offset, timeUnitsCount = 0, durationInTimeUnits;
+
+		dateFrom = new db.Date(location.query.get('dateFrom').value);
+		dateTo = query.dateTo || new db.Date();
+		mode = query.mode;
+		durationInTimeUnits = calculateDurationByMode(dateFrom, dateTo, mode);
+		pagination.count.value = Math.ceil(durationInTimeUnits / itemsPerPage);
+		pagination.current.value = Number(location.query.get('page').value) || 1;
+		if (pagination.count.value > 1) {
```
medikoo: - `dateTo`, `dateFrom` should already be final on `query` (no other processing of those values should be needed, apart of serializing them to ISO string before sending them with query to the server)
- Paginator interface should not be changed at this moment (we should do it when we fill table with data, line 178).
- `page` resolution should be made in queryHandler (as it happens in other paginated tables, e.g. business processes table). It's important that `queryHandler` validates complete query and not part of it (mind it also fixes the query in address bar if it's invalid, you don't want to pursue this things manually aside).

You can get notion of what handling is expected by investigating other configurations. e.g. in what we do in case of `inspector`:
  1. Query handler normalizes and validates query, after that's done it emits `query` event -> https://github.com/egovernment/eregistrations/blob/master/view/components/inspector-table/setup-query-handler.js#L14
  2. That runs `manager.update` which just does this -> https://github.com/egovernment/eregistrations/blob/master/view/components/objects-table-simple/manager.js#L46-L51. This is about some internal caching, and not really needed (at least at this point) normalization of `query.page` to number. `query` is passed to `queryExternal`.
  3. `queryExternal` directly passes query to server -> https://github.com/egovernment/eregistrations/blob/master/view/components/inspector-table/manager.js#L19-L20

Best if this configuration follows same convention.

The one tricky part, (but nothing that can't be cleanly solved), It's fact that for _paginator interface update_ (which should be done after we receive result from server) we need to know total number of pages, which is also internally needed by _query handler_ to properly resolve `page`.
I see few solutions:
- (better one) Introduce generic util that calculates `pagesCount` based on query, and for one update invoke it twice (1) internally in `queryHandler` and (2) when want to update paginator interface.
It seems not perfectly clean when we think of algorithm (same calculation runs twice), but we can easily fix that with memoization trick (this util will be run twice with same query object) in that case we can also rely on [weak map based memoization](https://github.com/medikoo/memoizee#weakmap-based-configurations) so we're friendly for garbage collector
- As we do not want to pass exactly same `query` to server, we can include `pagesCount` calculation logic purely in query handler, and from it expose both `page` and `pageCount` on resolved `query` (in query conf functions `ensure` receives `query` as second argument, so we can add `pageCount` in `page` handler).. Then from this `query` we can read values for paginator

kamsi:
@medikoo I did query handling of page in client. However for the `dateTo` I really need to modify it for building the server query, so that the server gives me the range which follows from pagination.

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,322 @@
...
+		date.setUTCDate(date.getUTCDate() + 1);
+		break;
+	}
+
+	return date;
+};
+
+var calculateDurationByMode = function (dateFrom, dateTo, mode) {
+	var timeUnitsCount = 0, currentDate;
+	currentDate = new db.Date(dateFrom.getTime());
```
medikoo: Unfortunately `getTime` is not reliable for copying `db.Date` that way, as it will resolve to local date in given stamp. Instead we should: `new db.Date(dateFrom.getUTCFullYear(), dateFrom.getUTCMonth().... )`

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,322 @@
...
+			durationInTimeUnits = calculateDurationByMode(dateFrom, dateTo, mode);
+			offset = { from: ((page - 1) * itemsPerPage) };
+			offset.to = offset.from;
+			if ((durationInTimeUnits - offset.from) < itemsPerPage) {
+				offset.to += durationInTimeUnits - offset.from;
+			} else {
+				offset.to += itemsPerPage;
+			}
+
+			currentDate = new db.Date(dateFrom.getTime());
```
medikoo: Here same way, to have accurate results we should _floor_ currentDate initially to given unit.

Otherwise in e.g. `month` mode, if dateFrom is 20th of some month, we will always ask for period starting or ending at 20th of some month, even when we ask for page 3 of 5.

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,322 @@
...
+			}
+		}
+		if (dateFrom) {
+			serverQuery.dateFrom = dateFrom.toJSON();
+		}
+
+		if (dateTo) {
+			serverQuery.dateTo = dateTo.toJSON();
+		}
+		queryServer(serverQuery).done(function (responseData) {
```
medikoo: Let's also before delete all query params not applicable to `serverQuery`

view/statistics-flow-certificates.js
medikoo: This is not addressed

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,312 @@
...
+							if (!data[key][service].certificate[certificate]) return;
+							rowsToAccumulate.push(data[key][service].certificate[certificate]);
+						} else {
+							rowsToAccumulate.push(data[key][service].businessProcess);
+						}
+					});
+					result[key] = accumulateResultRows(rowsToAccumulate);
+				}
+			});
+		});
```
medikoo: I take that server can send data with holes e.g. omit days for which it doesn't have data. So result can be:

```javascript
{
  '2016-01-23': ...
  '2016-01-27': ...
  '2016-02-01': ...
}
```

It would also be optimal for size of request.

In such case I believe we should fill missing dates with `buildResultRow()`

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,322 @@
...
+	modes.some(function (modeItem) {
+		if (modeItem.key === query.mode) {
+			mode = modeItem;
+			return true;
+		}
+	});
+
+	while (currentDate < dateTo) {
+		result[mode.getDisplayedKey(currentDate)] = buildResultRow();
+		incrementDateByTimeUnit(currentDate, mode.key);
```
medikoo: Same problem we have here, `currentDate` needs to be _floored_ up front, otherwise we do not guarantee expected incrementation

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	Object.keys(data).forEach(function (key) {
```
medikoo: I think it'll be clear to do upper `while` and this `forEach` in one go, as:

```javascript
while (...) {
  result[resolvedDisplayDate] = buildResultRow(resolvedDataFromResult);
}
```

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,322 @@
...
+  , serviceQuery      = require('../apps-common/query-conf/service')
+  , certificateQuery  = require('../apps-common/query-conf/certificate');
+
+exports._parent        = require('./statistics-flow');
+exports._customFilters = Function.prototype;
+
+exports['flow-nav']                = { class: { 'submitted-menu-item-active': true } };
+exports['flow-by-certificate-nav'] = { class: { 'pills-nav-active': true } };
+
+var queryServer = memoize(function (query) {
```
medikoo: As we intend to reuse same url across pages, let's seclude it into outer util so it can be used by further views

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,322 @@
...
+	rows.slice(1).forEach(function (row) {
+		Object.keys(row).forEach(function (propertyName) {
+			result[propertyName] += row[propertyName];
+		});
+	});
+
+	return result;
+};
+
+var copyDbDate = function (date) {
```
medikoo: This will serve well as `/utils/copy-db-date`

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,340 @@
...
+
+			currentDate = copyDbDate(dateFrom);
+			floorToUnit(currentDate, mode);
+			while (timeUnitsCount <= offset.to) {
+				if (timeUnitsCount === offset.from && query.page > 1) {
+					dateFrom = copyDbDate(currentDate);
+				}
+				if (timeUnitsCount === offset.to && query.page < query.pageCount) {
+					dateTo = copyDbDate(ceilToUnit(currentDate, mode));
+					if (dateTo > now) {
```
medikoo: What's the purpose of  adjustment in this condition? It doesn't seem needed

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,340 @@
...
+			offset.to -= 1;
+
+			currentDate = copyDbDate(dateFrom);
+			floorToUnit(currentDate, mode);
+			while (timeUnitsCount <= offset.to) {
+				if (timeUnitsCount === offset.from && query.page > 1) {
+					dateFrom = copyDbDate(currentDate);
+				}
+				if (timeUnitsCount === offset.to && query.page < query.pageCount) {
+					dateTo = copyDbDate(ceilToUnit(currentDate, mode));
```
medikoo: I think instead of introducing superfluous `ceilToUnit` we can just do:
```javascript
dateTo = incrementDateByTimeUnit(copyDbDate(currentDate), mode));
```

(there are also issues in `ceilToUnit`  but I didn't comment on them, as it's best if we remove it).

Additionally what's the agreement on `dateFrom` and `dateTo` handling? Are both inclusive?
If that's the case, then additionally we should:

```javascript
dateTo.setUTCDate(dateTo.getUTCDate() - 1);
```

(so e.g. instead of making `dateTo` the first day of the next month, we make it last day of current)

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,340 @@
...
+						dateTo.setUTCFullYear(now.getUTCFullYear());
+						dateTo.setUTCMonth(now.getUTCMonth());
+						dateTo.setUTCDate(now.getUTCDate());
+					}
+				}
+				timeUnitsCount++;
+				incrementDateByTimeUnit(currentDate, mode);
+			}
+		}
+		if (dateFrom) {
```
medikoo: I believe in all cases we have `dateFrom` and `dateTo`

test/utils/copy-db-date.js
```diff
@@ -0,0 +1,3 @@
+// TODO
```
medikoo: Let's add real tests here

utils/copy-db-date.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+var db = require('../db');
+
+module.exports = function (date) {
+	return new db.Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate());
```
medikoo: To avoid erroneous usage it might be good to require here instances of `db.Date` (best via `instanceof`)

It will also help to write accurate tests ;-)

test/utils/copy-db-date.js
```diff
@@ -0,0 +1,21 @@
...
+  , defineDbjsDate = require('dbjs-ext/date-time/date');
+
+module.exports = function (t, a) {
+	var date, dbDate;
+	defineDbjsDate(db);
+	date   = new db.Date();
+	dbDate = t(date);
+	date.setUTCFullYear(date.getUTCFullYear());
+	date.setUTCMonth(date.getUTCMonth());
+	date.setUTCDate(date.getUTCDate());
```
medikoo: What's the intention behind code in above three lines? It seems redundant, having zero effect

medikoo:
Other thing that I discovered now, is that we  still miss some additional logic to group result for a client. As `calculateStatusEventsSums` as being implemented at #1758 will answer with data needed for one table row.

We need some handler (to be invoked here: https://github.com/egovernment/eregistrations/blob/76fe43084d758b0141bfb573abc22d32d298d94e/server/routes/statistics.js#L74 ) that will seclude periods for each expected rows and call `calculateStatusEventsSums` with periods applicable for each row. So we send result as e.g.:

```
{
  "2016-01": periodSums, // Result of calculateEventsSum
  "2016-02": periodSums,
  "2016-03": periodSums
  etc.
}
```

kamsi:
@medikoo the utility for grouping the params for calls to `calculateEventsSum` is in: `utils/get-date-ranges-by-mode`. Since the last review I also included multiple other utils, which will be shared among views 1 and 2.

view/statistics-flow-certificates.js
```diff
@@ -0,0 +1,231 @@
...
+				timeUnitsCount++;
+				incrementDateByTimeUnit(currentDate, mode);
+			}
+		}
+		serverQuery.dateFrom = dateFrom.toJSON();
+		serverQuery.dateTo = dateTo.toJSON();
+
+		delete serverQuery.service;
+		delete serverQuery.page;
+		delete serverQuery.certificate;
```
medikoo: We also should now delete `pageCount`

server/routes/statistics.js
```diff
@@ -66,6 +69,12 @@ module.exports = function (config) {
 	getData(driver).done();
 
 	return assign({
+		'get-flow-data': function (query) {
+			return flowQueryHandler.resolve(query)(function (query) {
+				// data goes here
```
medikoo: I see utils secluded to external modules, but I don't see them used here to calculate needed periods (with which we will call `calculateStatusEventsSums` after #1758 is done)

mtuchowski:
Ready to merge to grouping PR https://github.com/egovernment/eregistrations/pull/1791

PR #1758, opened 2017-02-16
Stats: Tables: Periods calculation service

mtuchowski:
Needed for:

https://github.com/egovernment/eregistrations/issues/1756

https://docs.google.com/document/d/1i92UhTr_krhLYVc9TA88WGIrvwPAnM9x84WHCFpVtQc/edit

Service which prepares data for new stats.

Service should be a function which accepts the following:

`{ dateFrom: ..., dateTo:... }`

`dateFrom` and `dateTo` are optional.

By default the service should run its calculations for the day previous (relative to system timezone) to its calling. For example if called on 12-03-2017 it should calculate for 11-03-2017. However, if `dateFrom` was passed, the service should calculate result for each day between `dateFrom` and `dateTo` (if `dateTo` was not passed, then to the day previous to the day of execution).

The service needs to calculate all data that is needed for the new stats table for a given day. The data needs to be in such format that no further parsing is needed on the view (only summing up certain partial results is allowed). The calculated data is to be written to files in json format - one file per every day for which calculation was run. We can store the data in the `data-local/statistics` directory. The data calculated for a given day should be stored in a file names after the date, so if the calculation was run for 12-03-2017 the file should be `data-local/statistics/12-03-2017.json`.

The service needs to be run once every day at 2 A.M. of the time relative to systems time zone.
The service needs to be initiated by the main server process, so that it is run at the appropriate time.

@medikoo if you have an alternative solution, please add it here.


mtuchowski:
@medikoo OK, I finished. At least I don't see any more problems (it can be due to the fact that I've been looking for too long at the code ;) ).

One major thing to note is that I had to drop the live data update functionality in `getDateMap` in favour of keeping it in `getData` and doing cache clear every 24 hours (adjustable). This change is due to the fact that we wanted to cache the results of `calculateStatusEventsSums`. Caching the results there would make live update in `getDateMap` useless.

The way it's implemented is like so:

* `getData` handles all persistent storage data extraction and live update.
* `getDateMap` is memoized with maxAge set.
* `calculateStatusEventsSums` is also memoized with LRU (max 1000 entries) and maxAge set to the same value as `getDateMap`.

I think this is a pretty good setup and will prove satisfactory.

medikoo:
_dates map_ should not really be memoized, as it's more like log, where data is changing only in scope of current date (which we also never request in scope of statistics), it's natural no changes will occur in past dates, so no need for recalculation.

Intention was to have _dates map_ calculated once during db scan, and on the go gather data for day that is currently going (so at end of it we have complete information of it on _dates map_), no memoization or recalculation involved.

Lint is failing, so I'm sending it back

mtuchowski:
@medikoo I fixed the lint error and am sending this ticket back to review. Please look at the code (pointing out the errors of course) and (if possible) provide a note on how you see me changing the implemented behaviour because as of this moment I cannot see simple solutions.

server/business-process-query/get-data.js
```diff
@@ -125,24 +219,105 @@ module.exports = exports = memoize(function (driver) {
...
-				if (meta) {
-					if (meta.type && (meta.type !== 'direct')) return;
-					if (multiItemValue && !meta.multiple) return;
-					if (!meta.validate(record, multiItemValue)) return;
+				if (validateRecord(record, meta, multiItemValue)) {
 					meta.set(initBpDataset(bpId), record, multiItemValue);
+					return;
+				}
+				// Business process nested entities
+				if (exports.businessProcessNestedEntities.some(function (nestedEntityConf) {
```
medikoo: I see this generic map was made purely to gather statusHistory, it's not really needed. We also need to be careful what we do here, as performance and outcome of this scan has strong influence on time of system startup, and on consumed memory of a process.

Gathering `statusHistory` should be done with minimal effort and no trace of it should be left after we resolved _dates map_.

Therefore it definitely should not be mixed into data that we keep at `result.businessProcesses` and `result.steps`, those are entities that are meant to be kept forever in memory and serve very different purpose. Mixing those two different things into same data structures, makes things dirty

`statusHistory` data should be stored  on some other __temporary__ object (we may keep it e.g. on `result._statusHistory`), which we can get rid of easily afterward we have _dates map_ via e.g.:

```javascript
delete result._statusHistory
```

Data at `result._statusHistory` could be stored it form:

```javascript
businessProcesess[businessProcessId] = { status1Id: { stamp: .. , status: .., ..},  status2Id: { .. } }
processingSteps[stepId] = { status1Id: { stamp: .. , status: .., ..},  status2Id: { .. } }
certificates[certificateId] = { status1Id: { stamp: .. , status: .., ..},  status2Id: { .. } }
```
---

Therefore instead of doing here superfluous `exports.businessProcessNestedEntities.some`

We should just do regex match for status history event (same as e.g. we do for step keyPath: https://github.com/egovernment/eregistrations/blob/master/server/business-process-query/get-data.js#L135) which will confirm whether we deal with `statusHistory` record.

Also statusHistory records are not live, we do not need three-function handling via `validate`, `set`, `delete` which are dedicated for repetetive set/delete handling of same records.

It should be handled more straightforward way as e.g.

```javascript
var addStatusHistoryRecord = function (map, id, key, value) {
  if (!map[id]) map[id] = {};
  if (key === 'status') {
     if (value[0] !== '3') return;
     map[id].status = value;
  } else if (key === 'processor') {
     ...
  }
};

addStatusHistoryRecord(result._statusHistory.businessProcesses[businessProcessId],
statusHistoryId, statusHistoryKey, statusHistoryValue);
```

server/services/status-event-tracker.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+var getData    = require('../business-process-query/get-data')
+  , getDateMap = require('../business-process-query/get-status-history-date-map');
```
medikoo: We should not create a service for that, but initialize it in base utility that needs it.

e.g. `getData` is initialized in statistics routes  -> https://github.com/egovernment/eregistrations/blob/master/server/routes/statistics.js#L66

The `getDateMap` should be initialized most likely in body of `calculate-status-events-sums.js`

server/business-process-query/get-data.js
```diff
@@ -110,10 +138,76 @@ module.exports = exports = memoize(function (driver) {
 			});
 		});
 
+		storage.on('update', function (event) {
```
medikoo: This handling should be in `get-status-history-date-map.js ` and should update data on currentDate on _dates map_

server/business-process-query/get-status-history-date-map.js
```diff
@@ -0,0 +1,420 @@
...
+};
+ */
+
+var getStatusHistoryLogs = function (statusHistory) {
+	return Array.from(statusHistory.values()).sort(function (a, b) {
+		return a.date - b.date;
+	});
+};
+
+module.exports = memoize(function (data) {
```
medikoo: Idea of having all status histories at all times in process memory, and on 24 hours intervals calculating from it _dates map_ is not efficient. We should not be reckless with keeping data in memory which we do not need keep.

It was specified that data of status histories should be gathered once and discarded right after we calculate _dates map_. Then we should update data for current date on _dates map_ by observing events on storages.
Let's follow this.

if incrementing data in _dates map_ per incoming status history records seems for some reason difficult, or requires to write another sophisticated handler, there are possibly ways to have something reusable that could be used to build _date map_ same way for _past_ and _upcoming_ events.
We do not need to provide any data for day which hasn't ended, we may add it to _dates map_ after it's done. So e.g. we may have a dedicated function `generateDay` which will be called for all past dates at initialization, and then after every day ends we may call it for day that passed.

Also this module should rather get data on it's own and not be fed with it, so it should be more as:

```javascript
module.exports = getData(..)(function (data) {
 // generate and return dates map
});
```

server/services/calculate-status-events-sums.js
```diff
@@ -0,0 +1,274 @@
...
+
+	return getData(driver)(function (data) {
+		return calculate(getDateMap(data), data, dateFrom, dateTo);
+	}).aside(function () {
+		debugLoad('%s - %s calculate status events sums (in %s)', dateFrom, dateTo,
+			humanize(Date.now() - startTime));
+	});
+}, {
+	max: 1000,
+	maxAge: env.statisticsResolution || 1000 * 60 * 60 * 24, // 24 hours by default
```
medikoo: Results have no right to change, so time invalidation makes no sense

server/services/calculate-status-events-sums.js
```diff
@@ -0,0 +1,274 @@
...
+					oForEach(statusResult, eachStatusSetToSum);
+				});
+			});
+		});
+	});
+
+	return resultMap;
+};
+
+module.exports = memoize(function (dateFrom, dateTo) {
```
medikoo: `dateFrom` and `dateTo` should be here db.Date instances, I see in code that they're assumed as ISO strings, it should not be like that.

Also memoized function should not be asynchronous, it complicates memoization handling unnecessary, it would be nicer as e.g:

```javascript

var getCalculate = getDatesMap()(function (datesMap) {
   return memoize(function (dateFrom, dateTo) { .. }) ;
});

module.exports = function (dateFrom, dateTo) {
  reutrn getCalculate()(function (calculate) { return calculate(dateFrom, dateTo); });
};
```

medikoo:
>@vianneyl expressed that he would like to get that feature in ELS production today (I know that it's already late but as stated, couldn't get in touch earlier).

I think we should not follow:
- It touches core functionality that affects scalability of a system
- When we were starting work, it was estimated to @vianneyl that feature will ready at 3rd of March, assuming you will work with @kamsi on that, as @kamsi is off this week, this estimates to 10th of March, you ending it on your own. So we're not behind the schedule, and we should maintain natural pace.

Now let me review latest changes..

vianneyl:
@medikoo it is ok, we do not want to have this online today

2017-02-28 15:43 GMT+01:00 Mariusz Nowak <notifications@github.com>:

medikoo:
> The computational power to do that and to calculate for all dates is roughly the same. Either way I have to do a 'full db scan' because I need to have previous events for the calculations and not only last day's ones.

I take you need the previous events from `statusHistory` for which new event was added, but you shouldn't need anything else, right? If it so, It doesn't require full db scan. For each entity you can get it via way narrower query e.g.:

```javascript
// All statusHistory for businessProcess
storage.getObject(bpId + '/statusHistory')(function (allStatusHistoryRecords) {
});
```

So after end of a day, you know for which entities the new events came, and you can get other statusHistory events just specifically for those entities, and build datesMap just for this day.

Doing whole scan for that is not scalable. We should keep just __one__  only place where whole db scan, is done. It's in `getData`. If at some point it may appear as not scalable it we be doable to somehow workaround this __single__ scan, if we introduce full scan in other places in not organized way, we badly affect stability and maintanance of a system (e.g. on GT initial full scan in `getData` already takes 3 minutes, we should not do it anywhere else, even in 24 hour intervals)

server/business-process-query/get-data.js
```diff
@@ -125,24 +156,36 @@ module.exports = exports = memoize(function (driver) {
...
+					stepKeyPath = match[2];
+
+					if (!stepPaths.has(stepPath)) return;
+
+					// Processing step properties
+					meta = exports.stepMetaMap[stepKeyPath];
+					if (validateRecord(record, meta, multiItemValue)) {
+						meta.set(initStepDataset(stepPath, bpId), record, multiItemValue);
+						return;
+					}
```
mtuchowski: This resolver was moved to external utility that (as you noticed) does second db scan. This is to overcome one problem. To explain the problem let's do a quick recap:

We need to get `statusHistory` for `dateMap` computation. The code to get it was added to `getData` module. This caused the data to persist in memory which is not good (I agree with you on this one). You suggested to store it somewhere (`_statusHistory` on result from `getData`) that can be removed after the initial scan.

Now, `getData` can be called only once during the system lifetime, after that it keeps returning the same memoized result and updates it on the go. To properly compute the `dateMap` I need all `statusHistory` logs for each BP that had changes. The first time I need that, during a system startup, I would remove it after use from `getData` result to preserve memory. After the day passes, I'll need to get that logs for the second time. I cannot get it from `getData` as calling it for the second time will give me the result with `statusHistory` removed.

I opted to create a separate `statusHistory` resolver that does in fact do a separate db scan but only when needed and don't memoize the result. After `dateMap` calculation is complete this data is discarded preserving memory.

At the time of writing, scanning db for `statusHistory` in this separate resolver, together with computing `dateMap` takes around 1 or 2 minutes on my laptop using GT database. This is done once on system start and the at the end of each day. Also, this process does not stop the rest of app startup, it runs independently and only lags the first call to get the stats if user logs fast enough after the system start to request it (10 minutes after http server starts this is long over). So the problem with system startup time is marginal.

So, as explained above, I cannot have `statusHistory` resolver in `getData`.

medikoo: > To properly compute the dateMap I need all statusHistory logs for each BP that had changes.

Yes, I  explained in https://github.com/egovernment/eregistrations/pull/1758#issuecomment-283069864 how you can query for this data specifically without doing whole db scan, it's straightforward.

We need to maintain one db scan done at system initialization, and do not introduce any further scans over process run (`business-process-query` was introduced purely because [we had performance issues caused by many db scans](https://github.com/egovernment/eregistrations/pull/1602))

PR #1782, opened 2017-02-17
Add Close button for login/register modals in Demo

melux:
Login and register modals in demo mode miss close button, let's fix that.
Will fixe https://github.com/egovernment/eregistrations-salvador/issues/1718

### DIFF
--- a/view/components/login-dialog.js
+++ b/view/components/login-dialog.js
@@ -7,7 +7,7 @@ var db        = require('mano').db
 module.exports = dialog(
 	{ id: 'login', class: 'dialog-login dialog-modal' },
 	header(
-		a({ class: 'close', type: 'button', href: "" }, i({ class: 'fa fa-close' })),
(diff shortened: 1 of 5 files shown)


melux:
@medikoo ready for review, merge and update. Especially awaited on ELS and GT. Thanks!

view/components/login-dialog.js
```diff
@@ -7,6 +7,7 @@ var db        = require('mano').db
 module.exports = dialog(
 	{ id: 'login', class: 'dialog-login dialog-modal' },
 	header(
+		a({ class: 'close', type: 'button', href: "." }, i({ class: 'fa fa-close' })),
```
medikoo: Introduced class,  doesn't follow name conventions as described here: http://help.eregistrations.org/framework/styles/#class-naming-conventions

melux: yes renamed!

medikoo:
Merged into all systems

PR #1767, opened 2017-02-14
Stats: Tables: Roles: HTML view

kamsi:
Specification: https://docs.google.com/document/d/1i92UhTr_krhLYVc9TA88WGIrvwPAnM9x84WHCFpVtQc/edit

Additional important details to be found at #1759

UX mockup: http://xwv4ez.axshare.com/#p=table_of_quantities_of_files_in_roles__per_status

It should come with server GET controller that will retrieve needed data using utilities configured at #1758 

The filter bar should expose on url query with: `service`, `status`, `dateFrom`, `dateTo` and `mode` (either _daily_, _weekly_, _monthly_ or _yearly_).

View is expected to query server with `dateFrom`, `dateTo`,  `mode` (either _daily_, _weekly_, _monthly_ or _yearly_) and `page` params. The _service_ and _status_ filter should be resolved locally (server will respond with data for all combinations).
Result of server queries should be cached on client, we should not query server with same question twice (there's no chance result will change).

The data format as returned by server controller to client was already agreed at #1758, it'll be as:

for daily `mode`:

```javascript
{
  "2016-01-02": periodSums,
  "2016-01-03": periodSums,
  "2016-01-04": periodSums
  etc.
}
```
or for monthly `mode`:
```javascript
{
  "2016-01": periodSums, 
  "2016-02": periodSums,
  "2016-03": periodSums
  etc.
}
```

Where `periodSums` should most likely be reduced results of `calcuateStatusEventsSums` (as implemented at #1758)

The _direct_ result of `calcuateStatusEventsSums` would be as:

```
data.businessProcess[name][status] = sum
data.certificate[name][status] = sum
data.processingStep[name][processorId][status] = sum
```

While for this view we need to have it reduced to:

```
data.processingStep[name][status] = sum
```

The normalization should be done on server (so client receives it already normalized) as we will also need same data format for PDF and CSV views as generated on server.

PR #1770, opened 2017-02-20
Stats: Tables: Roles/Operators: HTML view

kamsi:
Specification: https://docs.google.com/document/d/1i92UhTr_krhLYVc9TA88WGIrvwPAnM9x84WHCFpVtQc/edit

Additional important details to be found at #1759

UX mockup: http://xwv4ez.axshare.com/#p=table_of_quantities_of_files_processed_in_roles__p

It should come with server GET controller that will retrieve needed data using utilities configured at #1758 

The filter bar should expose on url query with: `service`, `step`, `operator`, `dateFrom`, `dateTo` and `mode` (either _daily_, _weekly_, _monthly_ or _yearly_).
Additionally each roles will have individual pagination which will enrich url query with queries as `revisionPage`, `precalPage` etc.

Server should be queried with two kind of queries:
A. Data for all tables (strictly first page)
B. Data for specific table (any page apart of first).

When we enter the page, or change settings in main filter page for tables is alwasy reset to `1` (therefore all eventual `revisionPage`, `precalPage` etc. should dissapear from query).

In that case we will query server with query of kind _A_, in which we pass:  `dateFrom`, `dateTo`,  and `mode`.

In case when user operates with paginator of one of chosen tables, then we quer server with query of kind _B_, in which we pass: `step`, `dateFrom`, `dateTo`, `mode` and `page`

Otherwise _service_,  _status_ and _operator_ filters should be resolved locally (server will respond with data for all combinations).

Result of server queries should be cached on client, we should not query server with same question twice (there's no chance result will change).

The data format as returned by server controller to client was already agreed at #1758, it'll be as:

for daily `mode`:

```javascript
{
  "2016-01-02": periodSums,
  "2016-01-03": periodSums,
  "2016-01-04": periodSums
  etc.
}
```
or for monthly `mode`:
```javascript
{
  "2016-01": periodSums, 
  "2016-02": periodSums,
  "2016-03": periodSums
  etc.
}
```

Where `periodSums` should most likely be reduced results of `calcuateStatusEventsSums` (as implemented at #1758)

The _direct_ result of `calcuateStatusEventsSums` would be as:

```
data.businessProcess[name][status] = sum
data.certificate[name][status] = sum
data.processingStep[name][processorId][status] = sum
```

While for this view we need to have it reduced to:
In query of kind A:
```
data.processingStep[name][processorId][status] = sum
```
In query of kind B:
```
data.processingStep.stepName[processorId][status] = sum
```

The normalization best if done on server (so client receives it already normalized) as we will also need same data format for PDF and CSV views as generated on server.

### DIFF
--- a/apps/statistics/get-query-conf.js
+++ b/apps/statistics/get-query-conf.js
@@ -2,10 +2,7 @@
 'use strict';
 
 var aFrom              = require('es5-ext/array/from')
-  , ensureArray        = require('es5-ext/array/valid-array')
-  , normalizeOptions   = require('es5-ext/object/normalize-options')
-
-  , stringify          = JSON.stringify;
+  , ensureArray        = require('es5-ext/array/valid-array');
 
 var queryConf = [
 	require('../../apps-common/query-conf/date-from'),
@@ -13,22 +10,11 @@ var queryConf = [
 ];
 
 module.exports = exports = function (data) {
-	var options = normalizeOptions(data)
-	  , processingStepsMeta = options.processingStepsMeta
-	  , conf = aFrom(queryConf);
+	var conf = aFrom(queryConf);
 
 	conf.push(
 		require('../../apps-common/query-conf/service'),
-		{
-			name: 'step',
-			ensure: function (value) {
-				if (!value) return;
-				if (!processingStepsMeta[value]) {
-					throw new Error("Unrecognized step value " + stringify(value));
-				}
-				return value;
-			}
-		}
+		require('../../apps-common/query-conf/step')()
 	);
 
 	if (exports.customQueryConf) {
--- a/server/routes/statistics.js
+++ b/server/routes/statistics.js
@@ -96,7 +96,7 @@ module.exports = function (config) {
 				dateRanges = getDateRangesByMode(query.dateFrom, query.dateTo, query.mode);
 				dateRanges.forEach(function (dateRange) {
 					// dateRange: { dateFrom: db.Date, dateTo: db.Date } with dateRange query for result
-					result[mode.getDisplayedKey(dateRange.dateFrom)] = null;
+					result[mode.getDisplayedKey(dateRange.dateFrom)] = {};
 				});
 
 				result = reduceOperators(assign({ data: result }, query));
--- a/server/routes/statistics.js
+++ b/server/routes/statistics.js
@@ -25,6 +25,7 @@ var assign                  = require('es5-ext/object/assign')
   , getDateRangesByMode     = require('../../utils/get-date-ranges-by-mode')
   , modes                   = require('../../utils/statistics-flow-group-modes')
   , reduceOperators         = require('../../utils/statistics-flow-reduce-operators')
+  , itemsPerPage            = require('../../conf/objects-list-items-per-page')
   , flowQueryOperatorsHandlerConf = require('../../apps/statistics/flow-query-operators-conf');
 
 module.exports = function (config) {
@@ -89,15 +90,32 @@ module.exports = function (config) {
 		},
 		'get-flow-roles-operators-data': function (query) {
 			return flowQueryHandlerOperators.resolve(query)(function (query) {
-				var dateRanges, result = {}, mode;
+				var dateRanges, result = {}, mode, page = Number(query.page)
+				  , finalResult = {}, itemsCnt = 0, currentPage = 1;
 				mode = modes.get(query.mode);
 				dateRanges = getDateRangesByMode(query.dateFrom, query.dateTo, query.mode);
 				dateRanges.forEach(function (dateRange) {
 					// dateRange: { dateFrom: db.Date, dateTo: db.Date } with dateRange query for result
-					result[mode.getDisplayedKey(dateRange.dateFrom)] = {};
+					result[mode.getDisplayedKey(dateRange.dateFrom)] = null;
 				});
 
-				return reduceOperators(assign({ data: result }, query));
+				result = reduceOperators(assign({ data: result }, query));
+				Object.keys(result).forEach(function (date) {
+					Object.keys(result[date]).forEach(function (processorId) {
+						itemsCnt++;
+						if ((itemsCnt % itemsPerPage) === 1 && itemsCnt > 1) {
+							currentPage++;
+						}
+						if (currentPage === page) {
+							if (!finalResult[date]) {
+								finalResult[date] = {};
+							}
+							finalResult[date][processorId] = result[date][processorId];
+						}
+					});
+				});
+
+				return { data: finalResult, pageCount: currentPage };
 			});
 		},
 		'get-time-per-role': function (query) {
--- a/server/routes/statistics.js
+++ b/server/routes/statistics.js
@@ -23,7 +23,9 @@ var assign                  = require('es5-ext/object/assign')
   , getBaseRoutes           = require('./authenticated')
   , processingStepsMeta     = require('../../processing-steps-meta')
   , getDateRangesByMode     = require('../../utils/get-date-ranges-by-mode')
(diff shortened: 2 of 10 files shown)


utils/statistics-flow-reduce-operators.js
```diff
@@ -0,0 +1,83 @@
...
+		} else {
+			result[status] += row[status].businessProcess;
+			result.processed += row[status].businessProcess;
+		}
+	});
+
+	return result;
+};
+
+module.exports = function (params) {
```
medikoo: It would be cleaner if `data`  and `query` are taken as separate arguments (`data` is not part of a `query`), it will also make invocation of that util more natural

medikoo: Also let's document what actually this util  does, with example input and output

view/components/filter-bar/select-user.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
medikoo: Let's document what actually this util does, with examples of input and output

view/statistics-flow-operators.js
```diff
@@ -0,0 +1,161 @@
...
+						})
+					}, getStepLabelByShortPath(step));
+				})
+			),
+			selectPeriodMode(),
+			p({ class: 'submit' }, input({ type: 'submit' }))));
+
+	section(pagination);
+	section({ class: "section-primary" },
+		location.query.get('step').map(function (key) {
```
medikoo: `step` may not be in `query` and we should not show empty page in such case but first step.

We should draw it same way as other views, where we have one observable object, which is updated whenever query changes, and it's change invokes table re-render.

Relying on query observables as above is error-prone (we already observed that in previous views)

apps/statistics/get-query-conf.js
```diff
@@ -2,33 +2,19 @@
...
-			name: 'step',
-			ensure: function (value) {
-				if (!value) return;
-				if (!processingStepsMeta[value]) {
-					throw new Error("Unrecognized step value " + stringify(value));
-				}
-				return value;
-			}
-		}
+		require('../../apps-common/query-conf/step')()
```
medikoo: What about default step? We should pass it I believe

server/routes/statistics.js
```diff
@@ -78,12 +82,42 @@ module.exports = function (config) {
...
+				var dateRanges, result = {}, mode, page = Number(query.page)
+				  , finalResult = {}, itemsCnt = 0, currentPage = 1;
+				mode = modes.get(query.mode);
+				dateRanges = getDateRangesByMode(query.dateFrom, query.dateTo, query.mode);
+				dateRanges.forEach(function (dateRange) {
+					// dateRange: { dateFrom: db.Date, dateTo: db.Date } with dateRange query for result
+					result[mode.getDisplayedKey(dateRange.dateFrom)] = {};
+				});
+
+				result = reduceOperators(result, query);
```
medikoo: If I read correctly result is just hash of empty objects (as filled at line 99), and it is passed to `reduceOperators`, which per documentation expect object that's is result of `calculateStatusEventsSums`.
Will it be filled after #1758 is ready? If it's so, maybe let's put some TODO comment here, at least it's totally not clear to me.

utils/statistics-flow-reduce-operators.js
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * FROM:
+ *
+ * [date][serviceName].businessProcess[status] = num;
+ * [date][serviceName].certificate[name][status] = num;
+ * [date][serviceName].processingStep[stepName].pending.businessProcess = num;
+ * [date][serviceName].processingStep[stepName].pending.certificate[name] = num;
+ * [date][serviceName].processingStep[stepName].byProcessor[processorId][status].businessProcess =
+ * num;
+ * [date][serviceName].processingStep[stepName].byProcessor[processorId][status].certificate[name] =
```
medikoo: For clarity it would be good to omit in this list items which do take part in reduction

utils/statistics-flow-reduce-operators.js
```diff
@@ -0,0 +1,108 @@
...
+ * num;
+ *
+ * TO:
+ * [date][processorId] = {
+ *  processor: processorId,
+ *  date: "YYYY-MM-DD",
```
medikoo: Why do we provide `processor` and `date` on this object? This information is on keys

utils/statistics-flow-reduce-operators.js
```diff
@@ -0,0 +1,108 @@
...
+
+		if (service) {
+			reducedRows = reducedRows.map(function (reducedRow) {
+				return reducedRow[service];
+			});
+		} else {
+			reducedRows = Object.keys(reducedRows[0]).map(function (serviceName) {
+				return reducedRows[0][serviceName];
+			});
+		}
```
medikoo: Above 10 lines seem overly complex for what they do, Same can be expressed purely by:

```javascript
var identity = require('es5-ext/function/identity')
  , toArray = require('es5-ext/array/to-array')

...

reducedRows = service ? [data[date][service]] : toArray(data[date], identity);
```

And intention is much easier to read from above (so we have better maintanability)

utils/statistics-flow-reduce-operators.js
```diff
@@ -0,0 +1,108 @@
...
+		reducedRows = reducedRows.map(function (row) {
+			return row.processingStep[step].byProcessor;
+		});
+
+		reducedRows.forEach(function (reducedRow) {
+			if (processor) {
+				processorRow = reducedRow[processor];
+				if (isEmptyResult(processorRow)) return;
+				finalResult[date][processor] =
+					buildResultByProcessor(processorRow, certificate, processor, date);
```
medikoo: If same processor processes files from two services, wouldn't that instead of summing both, override result from first service with other?

utils/statistics-flow-reduce-operators.js
```diff
@@ -0,0 +1,108 @@
...
+			} else {
+				Object.keys(reducedRow).forEach(function (processorId) {
+					processorRow = reducedRow[processorId];
+					if (isEmptyResult(processorRow)) return;
+					finalResult[date][processorId] =
+						buildResultByProcessor(processorRow, certificate, processorId, date);
```
medikoo: Same here

apps/statistics/get-query-conf.js
kamsi: No, why? There was no default here before.

utils/statistics-flow-reduce-operators.js
kamsi: It's just for convenience, so after that I can do on view:

https://github.com/egovernment/eregistrations/pull/1770/files#diff-4ef00791749a97816c63d0d33d2eef63R140

apps/statistics/get-query-conf.js
medikoo: Ok, indeed it applies to view, where filter by step is optional.

utils/statistics-flow-reduce-operators.js
medikoo: This duplicates information which already thater and that makes result more heavy (we send it to client so better if response is light).

Additionally you have this information out of a box in a view, just use utils that make it friendly and readable:

```javascript
var toArray = require('es5-ext/object/to-array');
...
tbody(toArray(result, function (dateResult, date) {
  return toArray(dateResult, function (processorResult, processorId) {
    return tr(toArray(processorResult, function (data, key) {
      ...
    });
  });
}));
```

mtuchowski:
Ready to merge to grouping PR https://github.com/egovernment/eregistrations/pull/1791 after https://github.com/egovernment/eregistrations/pull/1767 is done.

PR #1765, opened 2017-02-22
 Stats: Tables: Certificates: PDF generation

kamsi:
Server-side generated PDF representation of view implemented at #1764 

It should generate same filtered version of view as one from which PDF request was initiated (query parameters needs to be passed to PDF url).

Still possibly we can resign from pagination in sense, that if view is on page 1, we only generate page 1. I take that it's nicer if PDF generates all rows for range and does not apply pagination as it's on view.


apps/statistics/flow-query-certificates-pdf-conf.js
```diff
@@ -0,0 +1,20 @@
...
+  , dateToParamConfig = require('../../apps-common/query-conf/date-to');
+
+module.exports = [
+	require('../../apps-common/query-conf/date-from'),
+	assign({},
+		dateToParamConfig,
+		{
+			ensure: function (value) {
+				if (!value) return new db.Date();
+				return dateToParamConfig.ensure(value);
```
medikoo: Why do we need such customization just for PDF case? Seems inconsistent. Is there any reason query handlers should be different for HTML, PDF and CSV?

utils/statistics-flow-certificates-filter-result.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
```
medikoo: All changes affecting `1764-stats-table-certificates-view` should be made on that branch and not here.

Otherwise when testing, #1764 we will not test final version, and it's not possible to reliably test and accept things gradually

apps/statistics/flow-query-certificates-pdf-conf.js
kamsi: Two things. The `dateTo` param doesn't have to be different, so yes we could use the same code as here in both cases to have default `dateTo`. As to the whole query handler, they do differ, as here we need to pass more params like `service` and `certficate` (in the regular view those two params are handled on the client, and are explicitly removed from server query). Here I would prefer to keep the filtering logic on the server, so that really the data that needs to be displayed is passed to pdf renderer.

medikoo: Yes, with `service` and `certificate` it's true, in that case we need to pass it to server, understood.
Still then probably we can build this conf on spot in controller, as we do in other cases where it's not reusable.

With `dateTo` I don't see any difference, so I think customizations is unjustified.

view/statistics-flow-certificates.js
```diff
@@ -133,6 +138,11 @@ exports['statistics-main'] = function () {
 					name: 'dateTo', value: location.query.get('dateTo') })
 			),
 			selectPeriodMode(),
+			div(
+				a({ class: 'users-table-filter-bar-print', href:
+					getDynamicUrl('/flow-certificates-data.pdf', { only: params }),
```
medikoo: We should not propagate `page` in query, and PDF should display all records from given period (not only the page that is presented in HTML view). Is that ensured?

kamsi: Yes, it is...

kamsi:
@medikoo I proposed to leave my work assigned to you as we don't know when it is going to be merged, and probably better if it hangs on you and maybe at @mtuchowski if we don't push it all until tomorrow...

medikoo:
@kamsi yes definitely, still it might be good if it lies on developer which will handle production data tests of it after all prior tickets are merged into `statistics2`.

After having that tests confirmed as passing, it'll be ready for merge.

Therefore as you're on a leave for next 3 weeks, and it is extremely unlikely we will move to this point before tomorrow, I'm assigning it to @mtuchowski

mtuchowski:
Ready to merge to grouping PR https://github.com/egovernment/eregistrations/pull/1792

PR #1768, opened 2017-02-24
Stats: Tables: Roles: PDF generation

kamsi:
Server-side generated PDF representation of view implemented at #1767 

It should generate same filtered version of view as one from which PDF request was initiated (query parameters needs to be passed to PDF url).

Still view current page setting should not be reflected, we should just output all rows from selected period range (it's fine that PDF will span multiple pages)


mtuchowski:
Ready to merge to grouping PR https://github.com/egovernment/eregistrations/pull/1792 after https://github.com/egovernment/eregistrations/pull/1765 is done.

PR #1766, opened 2017-02-24
Stats: Tables: Certificates:  CSV generation

kamsi:
Server-side generated CSV representation of view implemented at #1764

It should generate sheet of data representing same filtered version of view as one from which CSV request was initiated (query parameters needs to be passed to CSV url).

Still the page at which view is at is irrelevant, in all cases we should generate rows for whole selected period.


mtuchowski:
Ready to merge to grouping PR https://github.com/egovernment/eregistrations/pull/1793

PR #1772, opened 2017-02-24
Stats: Tables: Roles/Operators: CSV generation

kamsi:
Server-side generated CSV representation of view implemented at #1770 

It should generate sheet of data representing same filtered version of view as one from which CSV request was initiated (query parameters needs to be passed to CSV url).

Still the page at which view is at is irrelevant, in all cases we should generate rows for whole selected period.

CSV representation should follow description in that comment: https://github.com/egovernment/eregistrations/issues/1759#issuecomment-275636701

### DIFF
--- a/server/csv-renderers/statistics-flow-operators.js
+++ b/server/csv-renderers/statistics-flow-operators.js
@@ -1,8 +1,10 @@
 'use strict';
 
-var ensureObject = require('es5-ext/object/valid-object')
-  , modes        = require('../../utils/statistics-flow-group-modes')
-  , _            = require('mano').i18n.bind('Statistics operators csv');
+var ensureObject     = require('es5-ext/object/valid-object')
+  , deferred         = require('deferred')
+  , modes            = require('../../utils/statistics-flow-group-modes')
+  , _                = require('mano').i18n.bind('Statistics operators csv')
+  , usersFullNameMap = require('../lib/user-full-names-map');
 
 module.exports = function (result, config) {
 	ensureObject(config);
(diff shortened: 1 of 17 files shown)


server/csv-renderers/statistics-flow-operators.js
```diff
@@ -0,0 +1,39 @@
...
+  , _                = require('mano').i18n.bind('Statistics operators csv')
+  , usersFullNameMap = require('../lib/user-full-names-map');
+
+module.exports = function (result, config) {
+	ensureObject(config);
+	var data, mode;
+	mode = modes.get(config.mode || 'daily');
+	data = [
+		[
+			mode.labelNoun,
```
medikoo: This doesn't reflect expected format, as described in this comment -> https://github.com/egovernment/eregistrations/issues/1759#issuecomment-275636701

First column was supposed to state step, and then we should list data for all steps, one after another

medikoo:
@mtuchowski I'm assigning it to you, as @kamsi is on long vacations

mtuchowski:
Ready to merge to grouping PR https://github.com/egovernment/eregistrations/pull/1793 after https://github.com/egovernment/eregistrations/pull/1769 is done.

server/csv-renderers/statistics-flow-operators.js
mtuchowski: As per talk with @roxanadina this no longer applies: The CVS export should contain exactly the data shown on the page (same as PDF).

PR #1794, opened 2017-03-13
Webservice - catalogs export

roxanadina:
Open an endpoint with all enums used in the system

### DIFF
--- a/webservice/export-catalogs.js
+++ b/webservice/export-catalogs.js
@@ -18,13 +18,32 @@ module.exports = memoize(function (db) {
 		});
 		return catalogItem;
 	});
-	[].push.apply(catalogs, enums);
+	catalogs.push({ enums: enums });
 
-	// services
+	// services + its roles
+	var roles = [];
+	var bpRoles ={};
+
 	var services = db.BusinessProcess.extensions.toArray().map(function (bp) {
+		bp.prototype.processingSteps.map.forEach(function (step){
+			// if group - check the substeps
+			if (step.steps) {
+					step.steps.map.forEach(function (subStep) {
+						bpRoles[subStep.key] = { name: subStep.key, label: subStep.label };
+					});
+			} else {
+				bpRoles[step.key] = { name: step.key, label: step.label };
+			}
+		});
+	
 		return { name: bp.__id__, label: bp.prototype.label };
 	});
+	roles = Object.keys(bpRoles).map(function (key) {
+			return bpRoles[key];
+	});
+
 	catalogs.push({ services: services });
+	catalogs.push({ roles: roles });
 
 	// documents
 	var documents = db.Document.extensions.toArray().map(function (doc) {
--- a/webservice/export-catalogs.js
+++ b/webservice/export-catalogs.js
@@ -1,0 +1,62 @@
+'use strict';
+
+var  ensureDb = require('dbjs/valid-dbjs')
+  , ensureType = require('dbjs/valid-dbjs-type');
+
+// make it array, it is easier for strongly typed end part to interpret it
+var allCatalogs = [];
+
+var initCatalogs = function (db)  {
+	// ensure db
+	ensureDb(db);
+
+	var catalogs = [];
+	// all enums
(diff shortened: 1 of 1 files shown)


webservice/export-catalogs.js
```diff
@@ -0,0 +1,55 @@
...
+		catalogItem.options = extension.members.toArray().map(function (option) {
+			return { name: option, label: extension.meta[option].label };
+		});
+		return catalogItem;
+	});
+	[].push.apply(catalogs, enums);
+
+	// services
+	var services = db.BusinessProcess.extensions.toArray().map(function (bp) {
+		return { name: bp.__id__, label: bp.label };
```
mtuchowski: Labels are on `bp.prototype.label`.

webservice/export-catalogs.js
```diff
@@ -0,0 +1,55 @@
...
+	});
+	catalogs.push({ services: services });
+
+	// registrations
+	var registrations = db.Registration.extensions.toArray().map(function (reg) {
```
mtuchowski: We don't (most of the time) create a classes for registrations (for example https://github.com/egovernment/eregistrations-guatemala/blob/master/model/business-process-company/registrations/business-patent.js#L7-L9) so this array is empty in GT. They are, in that case, available under `db.Registration.instances` but their `__id__` is not human friendly (`BusinessProcessMandate#/registrations/map/mandateInscription`). Using `reg.key` will help but may cause duplicates in Salvador.

Using both methods, `extensions` and `instances` will also cause duplicates as for each class we will have it's instances for each BP that uses it.

This have to be rethought…

webservice/export-catalogs.js
```diff
@@ -0,0 +1,55 @@
...
+	catalogs.push({ registrations: registrations });
+
+	// documents
+	var documents = db.Document.extensions.toArray().map(function (doc) {
+		return { name: doc.__id__, label: doc.label };
+	});
+	catalogs.push({ documents: documents });
+
+	// costs
+	var costs = db.Cost.extensions.toArray().map(function (cost) {
```
mtuchowski: Same case as in registrations.

webservice/export-catalogs.js
```diff
@@ -0,0 +1,43 @@
...
+	var enums = db.StringLine.extensions.toArray().filter(function (extension) {
+		return extension.meta && extension.members;
+	}).map(function (extension) {
+		var catalogItem = { name: extension.__id__ };
+		catalogItem.options = extension.members.toArray().map(function (option) {
+			return { name: option, label: extension.meta[option].label };
+		});
+		return catalogItem;
+	});
+	[].push.apply(catalogs, enums);
```
mtuchowski: Let's maybe add those enum types under `catalogs.enums` in line with `catalogs.services` etc.

roxanadina: @mtuchowski  they all are enums for the end consumer
i don't think this new grouping brings any extra value

kamsi: @roxanadina maybe it does not bring much semantically, but still @mtuchowski has a point here. It's like we use namespace for our important groups (`services`, `documents` etc.) while flood the catalogs with all others directly. In the worst case scenario we can hit a name collision, when I for example have an enum entry `services`. So we should group them at least for the purpose of avoiding that (still there will be possibility of name collision between enum entries themselves).

roxanadina:
@kamsi, @mtuchowski 
enums added under "enums"
roles also added

webservice/export-catalogs.js
```diff
@@ -0,0 +1,63 @@
...
+			return { name: option, label: extension.meta[option].label };
+		});
+		return catalogItem;
+	});
+	catalogs.push({ enums: enums });
+
+	// services + roles
+	var services = [];
+	var roles = [];
+	var bpRoles = {};
```
kamsi: I really thinks we should not encourage programmers to identify processingSteps with roles. At some point we might be asked to expose roles, and it is something different than steps. Let's stick to processingSteps.

roxanadina: @kamsi  true! my bad

PR #1781, opened 2017-03-22
Webservice: BusinessProcess JSON Data export

kamsi:
Goal is to create a module that exports JSON Data for given `BusinessProcess` instance.

I propose to place this module at `server/webservice/export-business-process-data.js`

Below different data groups are listed, with notes on how to best approach their export

### 1. Form sections data

I believe it is desired to export data organized in meaningful way, so have separately grouped properties originating from _guide_, _data forms_, _submission forms_, _certificate data forms_ and _processing step data forms_.

Still it's to be decided, whether we want to additionally reflect internal organization of given context, e.g. whether we want to reflect internal organization of _data forms_ (export properties grouped by sections) or we want to export everything flat way in scope of it.

Reflecting sub sections may be helpful for other end, which might be interested in specific section of data as a whole, but downside is that if some property will be moved between sections (such reorganizations happen), then its address in JSON will change, that may also be highly undesirable for other end. @roxanadina do what you feel would work best.

#### 1.1 Method in which data should be exported

I propose to handle it in similar way we handle generation of JSON's for snapshots. So we have dedicated method on _section_ which returns export for given section.
One that handles snapshots is `toJSON` (it's controversial name as it's too generic for this use case). I would probably name one we need for webservice as `toWebserviceJSON` (but any better proposals are welcome).

Best way to follow this, is to investigate existing `toJSON` implementations, that prepare data exports for snapshots e.g.:
- https://github.com/egovernment/eregistrations/blob/master/model/lib/property-groups-process.js#L43-L49
- https://github.com/egovernment/eregistrations/blob/master/model/form-section.js#L258-L266 - https://github.com/egovernment/eregistrations/blob/master/model/form-section-group.js#L110-L124

And taking inspiration from that implement `toWebserviceJSON` variants

#### 1.2 What sections to retrieve

In below examples `data` object would be result of JSON export.

- Guide
```javascript
data.determinants = businessProcess.determinants.toWebserviceJSON();
```
- Data Forms
```javascript
data.dataForms = businessProcess.dataForms.toWebserviceJSON();
```
- Submission Forms
```javascript
data.submissionForms = businessProcess.submissionForms.toWebserviceJSON();
```
- Certificates data (only if in Part B and only for issued certificates)

```javascript
// "Overview section" in all cases
data.certificates[name].overviewSection = businessProcess.certificates.map[name].overviewSection.toWebserviceJSON();
// Data form only if we have some fields (Technically toWebserviceJSON() may return null then)
data.certificates[name].dataForm = businessProcess.certificates.map[name].dataForm.toWebserviceJSON();
```
However this actually should be achieved via:
```javascript
data.certificates[name] = businessProcess.certificates.map[name].toWebserviceJSON();
```
Which will also export other valuable flow info which is described in below points

- Processing steps data (only if in Part B, only for approved steps and only if `dataForm` is used in a that step (which can be detected by fact on wether it has some fields to serialize, and if there are none `toWebserviceJSON()` may be configured to return `null`)
```javascript
data.processingSteps[name].dataForm = data.processingSteps.map[name].dataForm.toWebserviceJSON();
```
Best if actually achieved via:
```javascript
data.processingSteps[name] = data.processingSteps.map[name]toWebserviceJSON();
```
Which will also export other valuable flow info which is described in below points

### 2. Uploaded files (documents)

As we agreed we will not export binaries but urls at which binaries can be downloaded.

Same way I propose to prepare `toWebserviceJSON` methods of which content can be inspired from existing `toJSON` methods, see e.g.:
- Single file toJSON: https://github.com/egovernment/eregistrations/blob/master/model/file.js#L24-L27
- RequirementUpload toJSON: https://github.com/egovernment/eregistrations/blob/master/model/requirement-upload.js#L119-L132

We will need it for:
- Requirement Uploads (only uploaded):
```javascript
data.requirementUploads = businessProcess.requirementUploads.toWebserviceJSON();
```
- Payment Receipt Uploads (only uploaded)
```javascript
data.paymentReceiptUploads = businessProcess.paymentReceiptUploads.toWebserviceJSON();
```
- Certificate uploaded files (only for issued)
It should be covered internally by:
```javascript
data.certificates[name] = businessProcess.certificates.map[name].toWebserviceJSON();
```
Which should be inspired out of https://github.com/egovernment/eregistrations/blob/master/model/document.js#L133-L164

### 3. Additional characteristics

We need to export information on:

- Requested registrations
```javascript
data.registrations = businessProcess.registrations.toWebserviceJSON()
```
`registrations.toWebserviceJSON`, may export just array of registration names, or maybe a map of registrations where for each registrations we provide an info which certificates resolve out of it e.g.:
```javascript
{
  nit: { certificates: ['nit'] },
  nrc: { certificates: ['nrc'] }
}
```
The latter approach will be more complete.

- Payable costs
```javascript
data.costs = businessProcess.costs.toWebserviceJSON()
```
`toWebserviceJSON` should resolve both result cost, as also list of eventual paymentReceiptUpload names, so we can map them to the costs

### Business Process flow information

- Status
```javascript
data.status = businessProcess.status
```
- Times of important events
`lastModified` returns microseconds timestamp in UNIX epox, possibly more friendly would be returning it in milliseconds or in date time ISO format as e.g. `2017-02-10T11:51:27.517Z` (can be obtained via `(new Date(stamp / 1000)).toISOString()`
```javascript
data.createdAt =  businessProcess.lastModified;
// If in Part B
data.submittedAt =  businessProcess.getOwnDescriptor('isSubmitted').lastModified;
// If Approved (finalized successfully)
data.approvedAt =  businessProcess.getOwnDescriptor('isApproved').lastModified;
// If Rejected
data.rejectedAt =  businessProcess.getOwnDescriptor('isRejected').lastModified;
```

Same way we may need above info per processing step, in such case it should be covered at:

```javascript
data.processingSteps[name] = data.processingSteps.map[name].toWebserviceJSON()
```

---

Finally agreed data structure, should also be described using [JSON Schema](http://json-schema.org/) standard (if possible), and that's to be handled at #1780


model/base.js
```diff
@@ -54,6 +72,19 @@ module.exports = memoize(function (db) {
 				return value.toString(descriptor);
 			}
 			return (new this(value)).toString(descriptor);
+		} },
+		valueToWebServiceJSON: { type: db.Function, value: function (value, descriptor) {
+			var db = this.database;
+			if (value == null) return value;
+			// enums handling
+			if (this.meta && this.members) {
```
kamsi: Yes, I need to override `valueToWebServiceJSON` of Economic activity objects.

model/document.js
```diff
@@ -129,6 +129,25 @@ module.exports = memoize(function (db) {
...
 		} },
+		toWebServiceJSON: {
+			value: function (ignore) {
+				var data = {
+					code: this.uniqueKey,
+					files: [],
+					data: null
+				};
+				if (this.dataForm.constructor !== this.database.FormSectionBase) {
+					data.data = this.dataForm.toWebServiceJSON();
```
kamsi: Yes, not a big problem. There is a chance. We can have file type nested on any object. I will cut files from `dataForm` of documents only for now.

model/requirement-upload.js
```diff
@@ -116,6 +116,11 @@ module.exports = memoize(function (db) {
 			multiple: true
 		},
 
+		toWebServiceJSON: {
+			value: function (ignore) {
+				return this.document.toWebServiceJSON();
```
kamsi: Not really doable to provide owner, I think. I don't see what would qualify something for an owner. The closest I came by is such a parent object which is instance of Person class... but this is still really bad condition.

kamsi: I came up with something after all, should be fine.

model/document.js
```diff
@@ -129,6 +129,38 @@ module.exports = memoize(function (db) {
 		isReleased: { type: db.Boolean, value: function () {
 			return this.status === 'approved';
 		} },
+		toWebServiceJSON: {
+			value: function (ignore) {
+				var data = {
+					code: this.key,
```
roxanadina: this is "document" in uploads and payments cases:
https://gist.github.com/roxanadina/d8b5c7619f84fedea835386a7d2511f2#file-new-bp-json-L76

roxanadina:
@kamsi and if there's a easy way to apply the "prettify" mechanism in 
documents :)
https://gist.github.com/roxanadina/d8b5c7619f84fedea835386a7d2511f2#file-new-bp-json-L249
and 
https://gist.github.com/roxanadina/d8b5c7619f84fedea835386a7d2511f2#file-new-bp-json-L95

roxanadina:
@kamsi  - the only comment I still have is about processingStep status being inside a status prop https://gist.github.com/roxanadina/d8b5c7619f84fedea835386a7d2511f2#file-new-bp-json-L9

model/business-process-new/base.js
```diff
@@ -79,7 +79,88 @@ module.exports = memoize(function (db/*, options*/) {
...
+				});
+				this.submissionForms.applicable.forEach(function (section) {
+					dataFields = section.toWebServiceJSON();
+					Object.keys(dataFields).forEach(function (fieldName) {
+						result.request.data[fieldName] = dataFields[fieldName];
+					});
+				});
+
+				if (!Object.keys(result.request.data).length) {
+					result.request.data = null;
```
mtuchowski: Why in case of `data` you set it to `null` instead of just leaving it as empty hash like in case of other properties you allow empty arrays?

model/business-process-new/base.js
```diff
@@ -79,7 +79,88 @@ module.exports = memoize(function (db/*, options*/) {
...
+					result.processingSteps[processingStep.key] = processingStep.toWebServiceJSON();
+				});
+
+				this.registrations.requested.forEach(function (reg) {
+					result.request.registrations.push({
+						code: reg.key
+					});
+				});
+
+				this.requirementUploads.applicable.forEach(function (upload) {
```
kamsi: Is it sth more then a gut feeling?  I believe that applicable collection works fine, I checked it for ELS and got requirementUploads of nested objects.

model/business-process-new/base.js
```diff
@@ -79,7 +79,88 @@ module.exports = memoize(function (db/*, options*/) {
...
+					},
+					submittedTimestamp:
+						this._isSubmitted.lastModified ?
+								Math.floor(this._isSubmitted.lastModified / 1000) : null,
+					processingSteps: {},
+					request: {
+						registrations: [],
+						documentUploads: [],
+						costs: [],
+						payments: [],
```
kamsi: I don't think we have to use our internal name here, payments seems a better (and suggested in doc) name for this functionality.

model/business-process-new/base.js
```diff
@@ -79,7 +79,88 @@ module.exports = memoize(function (db/*, options*/) {
...
+					service: {
+						code: this.constructor.__id__
+					},
+					submittedTimestamp:
+						this._isSubmitted.lastModified ?
+								Math.floor(this._isSubmitted.lastModified / 1000) : null,
+					processingSteps: {},
+					request: {
+						registrations: [],
+						documentUploads: [],
```
kamsi: @roxanadina what do you say to this?

model/business-process-new/base.js
kamsi: `payments` sound better... @roxanadina what do you think?

model/business-process-new/base.js
roxanadina: @mtuchowski @kamsi 
I agree with kamsi, this is anyway hardcoded and it's a more "human" name
+ we already implemented it like this - in schema also :)

model/business-process-new/base.js
roxanadina: @mtuchowski @kamsi - same arguments as previous :)

PR #1799, opened 2017-03-23
Stats - times, include working hours calculation 

kamsi:
Write processing times in scope of statusHistory of steps.


kamsi:
@mtuchowski please don't merge yet. It is probable that there will be adjusts to this as the working hours will need to be set  in the end systems.

kamsi:
@mtuchowski I just got informed that we can use 8 - 16 as working hours for both GT and ELS, so no adjusts should be needed. Go ahead if you feel like merging...

utils/get-processing-working-hours-time.js
```diff
@@ -0,0 +1,66 @@
...
+	}
+	workTimePerDay = (workingHours.end.hours - workingHours.start.hours) * 3600000;
+	workTimePerDay += ((workingHours.end.minutes - workingHours.start.minutes) * 60000);
+
+	startDate = new db.Date(startDateTime.getFullYear(), startDateTime.getMonth(),
+		startDateTime.getDate());
+	endDate   = new db.Date(endDateTime.getFullYear(), endDateTime.getMonth(), endDateTime.getDate());
+
+	startDateTime.setHours(Math.max(startDateTime.getHours(), workingHours.start.hours));
+	startDateTime.setMinutes(startDateTime.getHours() < workingHours.end.hours ?
```
mtuchowski: For startDateTime: 12:15 and workingHours.end: 12:30 this will set 30 and not 15 as it should.

utils/get-processing-working-hours-time.js
```diff
@@ -0,0 +1,66 @@
...
+			startDateTime.getMinutes() : workingHours.end.minutes);
+
+	endDateTime.setHours(Math.min(endDateTime.getHours(), workingHours.end.hours));
+	endDateTime.setMinutes(endDateTime.getHours() < workingHours.end.hours ?
```
mtuchowski: Same error as with `startDateTime` above.

kamsi: I believe this was actually handled at the end...

mtuchowski:
While testing this on ELS I found it's not counting times for steps that took under an hour. For example, created file entered revision on `1490777822312223` and was approved on `1490778134553906`. The `processingWorkingHoursTime` was not updated.

PR #1790, opened 2017-03-23
Uploads - do not try to create preview for doc & xls files

raunokulla:


kamsi:
@raunokulla have you tested it in some end system?

raunokulla:
yes. Against Salvador's.

PR #1803, opened 2017-03-27
Modify component totals-in-matrix.js to skip empty paths and vertical total

raunokulla:
Existing component totals-in-matrix.js has to be changed in order to meet new requirements. Two changes have to be implemented.

Firstly, this component has to be able to configure so that it would not calculate vertical totals.

Secondly, all the properties do not have both asset and liability value, therefore it is necessary to change the component in the manner that in case the property path is empty, it will display empty nothing in the field position.

### DIFF
--- a/view/components/utils/totals-in-matrix.js
+++ b/view/components/utils/totals-in-matrix.js
@@ -8,7 +8,7 @@ var generateTableBody = function (rows, master) {
 	return rows.map(function (row) {
 		return tr(th(row.label, span({ class: 'hint' }, row.inputHint)),
 			list(row.paths, function (path) {
-				if (!path) return td();
+				if (path == null) return td();
 				var resolved = master.resolveSKeyPath(path);
 				return td(input({
 					control: { id: 'matrix-display-input-' + path },
@@ -72,12 +72,12 @@ module.exports = function (config) {
 			td({ id: totalId })
 		)),
 
-		script(function (formId, horizontalTotals, verticalTotals, skipVerticalTotal, totalId) {
+		script(function (formId, horizontalTotals, verticalTotals, totalId) {
 			var form = $(formId), totals = [], fullTotal = { total: $(totalId), subTotals: [] };
 			$.forEach(horizontalTotals.concat(verticalTotals), function (total) {
 				var subTotatals = [];
 				$.forEach(total.paths, function (path) {
-					if (!path) return;
+					if (path == null) return;
 					subTotatals.push($('matrix-display-input-' + path));
 				});
 				totals.push({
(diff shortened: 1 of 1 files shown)


view/components/utils/totals-in-matrix.js
```diff
@@ -8,6 +8,7 @@ var generateTableBody = function (rows, master) {
 	return rows.map(function (row) {
 		return tr(th(row.label, span({ class: 'hint' }, row.inputHint)),
 			list(row.paths, function (path) {
+				if (!path) return td();
```
kamsi: Let's rely on agreed explicit `null` comparison here (sloppy one):

`if (path == null) return td();`

We should treat it as contract, other falsy values like `''` or `false`, should not yield this result and should be treated as accidental errors.

view/components/utils/totals-in-matrix.js
```diff
@@ -68,11 +72,12 @@ module.exports = function (config) {
 			td({ id: totalId })
 		)),
 
-		script(function (formId, horizontalTotals, verticalTotals, totalId) {
+		script(function (formId, horizontalTotals, verticalTotals, skipVerticalTotal, totalId) {
 			var form = $(formId), totals = [], fullTotal = { total: $(totalId), subTotals: [] };
 			$.forEach(horizontalTotals.concat(verticalTotals), function (total) {
 				var subTotatals = [];
 				$.forEach(total.paths, function (path) {
+					if (!path) return;
```
kamsi: Same as above.

view/components/utils/totals-in-matrix.js
```diff
@@ -83,11 +88,12 @@ module.exports = function (config) {
 			$.forEach(verticalTotals, function (total) {
 				var subTotatals = [];
 				$.forEach(total.paths, function (path) {
+					if (!path) return;
 					subTotatals.push($('matrix-display-input-' + path));
 				});
 				fullTotal.subTotals = fullTotal.subTotals.concat(subTotatals);
 			});
-			totals.push(fullTotal);
+			if (!skipVerticalTotal) totals.push(fullTotal);
```
kamsi: So, if I see correctly, whenever I want to skipVerticalTotal, I will also skip total. This is weird, I thought you only wanted to skip vertical total not the whole total.

I believe this line is not needed because:

1. If skipVerticalTotal === true, then my verticalTotals is equal [] (so I can safaly update totals with fullTotal, as fullTotal may only have anything interesting in horizontalTotals).
2. If skipVerticalTotal is falsy, then it will work exactly as before this change.

Since you make sure that `skipVerticalTotal` will result in empty array passed as `verticalTotals`, then I don't think you even have to pass `skipVerticalTotal` to `$.onEnvUpdate`.

PR #1804, opened 2017-03-27
JQuery and Period selector integration

ilja903:


ilja903:
Integrated jquery, jquery ui, moment.js and datepicker. Note that this commit comes in pair with https://github.com/egovernment/eregistrations-salvador/pull/1782 (at first you have to accept one then I will have to update eregistrations core)

However, nothing bad will happen if you accept this one due to fact that I have implemented fallback mechanism: if there is no jquery it will show ordinary dateFrom and dateTo fields.

Unfortunately I had to ignore the linting of some files because they became too long to parse.

For testing port see: https://unctad.atlassian.net/wiki/display/EPC/Period+selector

view/components/filter-bar/select-date-range.js
```diff
@@ -0,0 +1,153 @@
...
+			var range = elem.daterangepicker("getRange");
+			jQuery('#startId').val(dateToString(range.start));
+			jQuery('#endId').val(dateToString(range.end));
+			$.dispatchEvent($('startId'), 'change');
+			$.dispatchEvent($('endId'), 'change');
+		});
+	});
+
+	return div(
+		input({
```
kamsi: This looks like mostly copy pasted version of existing `dateFrom` input builder. I think we could reuse util here. In case some options are missing (like `hidden` attribute, you can add them to the original util).

view/components/filter-bar/select-date-range.js
```diff
@@ -0,0 +1,153 @@
...
+			name: 'dateFrom',
+			value: location.query.get('dateFrom').map(function (dateFrom) {
+				if (dateFrom) return dateFrom;
+				var now = new db.Date(), defaultDate =
+					new db.Date(now.getUTCFullYear(), now.getUTCMonth(),
+						now.getUTCDate() - 6);
+				return defaultDate.toISOString().slice(0, 10);
+			})
+		}),
+		input({
```
kamsi: Same here with `dateFrom` being the source.

css/base.css
```diff
@@ -238,3 +238,132 @@ Styles applied while screen size is less than 640px
...
...osition:0 -208px}.ui-icon-circlesmall-minus{background-position:-16px -208px}.ui-icon-circlesmall-close{background-position:-32px -208px}.ui-icon-squaresmall-plus{background-position:-48px -208px}.ui-icon-squaresmall-minus{background-position:-64px -208px}.ui-icon-squaresmall-close{background-position:-80px -208px}.ui-icon-grip-dotted-vertical{background-position:0 -224px}.ui-icon-grip-dotted-horizontal{background-position:-16px -224px}.ui-icon-grip-solid-vertical{background-position:-32px -224px}.ui-icon-grip-solid-horizontal{background-position:-48px -224px}.ui-icon-gripsmall-diagonal-se{background-position:-64px -224px}.ui-icon-grip-diagonal-se{background-position:-80px -224px}.ui-corner-all,.ui-corner-top,.ui-corner-left,.ui-corner-tl{border-top-left-radius:3px}.ui-corner-all,.ui-corner-top,.ui-corner-right,.ui-corner-tr{border-top-right-radius:3px}.ui-corner-all,.ui-corner-bottom,.ui-corner-left,.ui-corner-bl{border-bottom-left-radius:3px}.ui-corner-all,.ui-corner-bottom,.ui-corner-right,.ui-corner-br{border-bottom-right-radius:3px}.ui-widget-overlay{background:#aaa;opacity:.003;filter:Alpha(Opacity=.3)}.ui-widget-shadow{-webkit-box-shadow:0 0 5px #666;box-shadow:0 0 5px #666}
```
mtuchowski: Can we move this to own css file? This would need adjusts in all end systems but I think it's worth it.

css/base.css
```diff
@@ -238,3 +238,132 @@ Styles applied while screen size is less than 640px
...
+
+/*!
+* Copyright (c) 2017 Tamble, Inc.
+* Licensed under MIT (https://github.com/tamble/jquery-ui-daterangepicker/raw/master/LICENSE.txt)
+*/
+
+.comiseo-daterangepicker-triggerbutton.ui-button {
```
mtuchowski: And this one also to a separate css file.

mtuchowski:
* Please try to align other elements in the same horizontal line:

    screenshot_2017-04-06_13-16-48
    screenshot_2017-04-06_13-16-58


* I don't know what's happening with that url:

    ```
    http://ilja9032-els.eregistrations.org/flow/?dateFrom=2017-01-18&dateTo=2017-01-22&=%7B%22start%22%3A%222017-01-18%22%2C%22end%22%3A%222017-01-22%22%7D&mode=daily
    ```

    Make sure that there is no garbage in it (`&=%7B%22start%22%3A%222017-01-18%22%2C%22end%22%3A%222017-01-22%22%7D`)

view/components/filter-bar/select-date-from.js
```diff
@@ -8,8 +8,8 @@ module.exports = function (/* opts */) {
 	name = opts.name || 'dateFrom';
 	return input({
 		id: opts.id || 'date-from-input',
-		type: 'date',
-		name: name,
+		type: opts.type || 'date',
+		name: opts.name || name,
```
kamsi: Just saw this one, it is redundant. In line 8 you already use opts.name to setup name.

mtuchowski:
Statistics got a new view, Reasons of rejection. You need to update it for new period selector:

screenshot_2017-04-11_10-23-05

PR #1807, opened 2017-03-29
Default setup of date selector

kamsi:
In all tables of the stats, we would like to have the same period selected by default. This period is from 01/01 of the current year to today (or to yesterday if there is no data for today).


mtuchowski:
Merge required.

PR #1809, opened 2017-04-03
Part B: Allow business processes without uploads.

mtuchowski:
This is in response to https://github.com/egovernment/eregistrations-salvador/pull/1737 that allow this situation.

### DIFF
--- a/routes/user.js
+++ b/routes/user.js
@@ -2,10 +2,11 @@
 
 'use strict';
 
-var hyphenToCamel    = require('es5-ext/string/#/hyphen-to-camel')
-  , db               = require('mano').db
-  , matchUpload      = require('./utils/user-match-upload')
-  , matchCertificate = require('./utils/user-match-certificate');
+var hyphenToCamel               = require('es5-ext/string/#/hyphen-to-camel')
+  , db                          = require('mano').db
+  , matchUpload                 = require('./utils/user-match-upload')
+  , matchFirstRequirementUpload = require('./utils/user-match-first-requirement-upload')
+  , matchCertificate            = require('./utils/user-match-certificate');
 
 var matchBusinessProcess = function (businessProcessId) {
 	this.businessProcess = db.BusinessProcess.getById(businessProcessId);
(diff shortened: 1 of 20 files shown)


routes/user.js
```diff
@@ -32,9 +33,7 @@ module.exports = {
 	'business-process/[0-9][a-z0-9]+/documents': {
 		match: function (businessProcessId) {
 			if (!matchBusinessProcess.call(this, businessProcessId)) return false;
-			var firstUpload = this.businessProcess.requirementUploads.dataSnapshot.resolved[0];
-			if (!firstUpload) return false;
-			return matchUpload.call(this, 'requirementUpload', firstUpload.uniqueKey);
+			return matchFirstRequirementUpload.call(this);
```
kamsi: No payment receipts nor certificates possible here?

view/business-process-official.js
```diff
@@ -11,11 +11,17 @@ exports._match = 'businessProcess';
 exports['sub-main'] = {
 	class: { content: true, 'user-forms': true },
 	content: function () {
+		var hasDocuments = or(
+			this.businessProcess.requirementUploads.applicable._size,
+			this.businessProcess.paymentReceiptUploads.applicable._size,
+			this.businessProcess.certificates.userApplicable._size
```
kamsi: I'm not sure that we should include `userApplicable` collection for certificates in context of official. I think that for official it should be `applicable`.

view/business-process-revision.js
```diff
@@ -168,64 +168,87 @@ exports._pauseButton = function (/*options*/) {
...
 exports._revisionContainer = function () {
-	var paymentUploads = getUploads(this.processingStep.paymentReceiptUploads, this.appName);
-	section({ class: 'section-tab-nav' },
-		a({
+	var requirementUploads    = this.processingStep.requirementUploads
+	  , paymentReceiptUploads = this.processingStep.paymentReceiptUploads
+	  , uploads               = getUploads(requirementUploads, this.appName)
+	  , paymentUploads        = getUploads(paymentReceiptUploads, this.appName);
+
+	var paymentUploadsTabNumber = resolve(uploads, '_length').map(function (uploadsLength) {
```
kamsi: Let's keep `paymentUploadsTabNumber` as an observable which stores a number. We should have observable for each tab number. And they should actually store numbers. One should calculate it's number as previous + 1. Then, once you pass those observables to i18n placeholders use `map` to transform the number value to string.

routes/user.js
mtuchowski: Nope, never was.

PR #1780, opened 2017-04-04
Webservice: BusinessProcess JSON Schema export

roxanadina:
Goal is to create a module that exports JSON schema for given `BusinessProcess` class. This schema will be important for end client to clarify semantics of JSON data we provide for specific business process instances.

I think nicest would be to rely on [JSON Schema](http://json-schema.org/) standard, assuming it's possible for our use case.

I propose to place this module at `server/webservice/export-business-process-schema.js`

The schema should describe data final data format on which we settled at #1781


roxanadina:
@kamsi , @mtuchowski the throne needs confirmation!

webservice/utils/get-data-from-form-section.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
+
+var assign = require('es5-ext/object/assign');
+
+module.exports = function getDataFromSection(dataForm, result, typeMapper) {
```
kamsi: OK, so we have `getDataFromSection` in the module which is called `get-data-from-form-section`... so which is it `FromSection` or `FromFormSection`?

webservice/utils/get-data-from-form-section.js
```diff
@@ -0,0 +1,63 @@
...
+	var ownerType = dataForm.propertyMaster;
+
+	// first try toWsSchemaJson method
+	if (dataForm.toWebServiceSchemaJSON) {
+		var dataFormResult = dataForm.toWebServiceSchemaJSON(typeMapper);
+		return assign(result, dataFormResult);
+	}
+
+	// if tabular section
+	if (dataForm.propertyName) {
```
kamsi: This is some serious duck typing, not really safe. Better check for instance of `FormEntitiesTable`. Still I don't see why `FormEntitiesTable` can't have `toWebServiceSchemaJSON`?

webservice/utils/get-data-from-form-section.js
```diff
@@ -0,0 +1,63 @@
...
+
+		// exclude customs
+		if (!ownerType.get(dataForm.propertyName)
+				|| !ownerType.get(dataForm.propertyName).getItemType) {
+			return;
+		}
+
+		var formOwner = ownerType.get(dataForm.propertyName).getItemType();
+		if (formOwner.prototype.dataForms && formOwner.prototype.dataForms.size) {
```
kamsi: This check is redundant.

webservice/utils/get-data-from-form-section.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
+
+var assign = require('es5-ext/object/assign');
+
+module.exports = function getDataFromSection(dataForm, result, typeMapper) {
+
+	if (!result) result = {};
+
+	var ownerType = dataForm.propertyMaster;
```
kamsi: What you call `ownerType` is not really a type but a prototype. I think the name is misleading.

webservice/utils/get-data-from-form-section.js
```diff
@@ -0,0 +1,63 @@
...
+	if (dataForm.toWebServiceSchemaJSON) {
+		var dataFormResult = dataForm.toWebServiceSchemaJSON(typeMapper);
+		return assign(result, dataFormResult);
+	}
+
+	// if tabular section
+	if (dataForm.propertyName) {
+
+		// exclude customs
+		if (!ownerType.get(dataForm.propertyName)
```
kamsi: What you really want to do here is check if `!ownerType.get(dataForm.propertyName) instanceof db.NestedMap` and it is enough. First of all `ownerType.get(dataForm.propertyName)` is true for proper configuration of section. Secondly `getItemType` is just duck typing for `NestedMap`.

webservice/utils/get-data-from-form-section.js
```diff
@@ -0,0 +1,63 @@
...
+				result[dataForm.propertyName] = {};
+				result[dataForm.propertyName].type = "array";
+				result[dataForm.propertyName].items = { type: "object", properties: { } };
+
+				getDataFromSection(subForm, result[dataForm.propertyName].items.properties, typeMapper);
+
+			});
+		}
+	} else {
+		(dataForm.propertyNames || []).forEach(function (prop) {
```
kamsi: `dataForm.propertyNames` is redundant unless you plan for some `FormSection` not to have `toWebServiceSchemaJSON` - which would be rather bad idea. If we take this situation out of the picture you are left no way `dataForm.propertyNames` being defined. Anyway `dataForm` is rather not the best choice of name - it should rather be `section`.

roxanadina:
note to self - remove required attribute from data fields

roxanadina:
additionally, we also need a dataForms array property at each `properties` level, that will allow grouping of properties per form if needed. 

the format should be` dataForms: [ { title: "", properties: []}]`

every entity that holds a dataForm should have this property (for example, nested branches dataForms will be inside its definition, at this level https://gist.github.com/roxanadina/c7b158a0336189845340ad667f7e7911#file-schema-els-json-L897 and not main level )
@mtuchowski  - correct me if I am wrong
@raunokulla - ask me or marek  if any questions

model/document.js
```diff
@@ -197,7 +197,32 @@ module.exports = memoize(function (db) {
...
+							data: {}
+						}
+					}
+				};
+				schema.items.properties.files =
+					this.files.map._descriptorPrototype_.type.prototype.toWSSchema();
+				schema.items.properties.owner = { type: "string" };
+				if (this.dataForm.constructor !== this.database.FormSectionBase) {
+					schema.items.properties.data = this.dataForm.toWSSchema();
+					schema.items.dataForms = [ { title: "", properties: [] } ];
```
kamsi: Looks a bit odd. Is this an attempt to get information about section in the schema? If, so then we should either fill it in with sth interesting, or remove this altogether.

model/document.js
```diff
@@ -197,7 +197,32 @@ module.exports = memoize(function (db) {
...
+							code: {
+								type: "enum",
+								ref: "documents"
+							},
+							data: {}
+						}
+					}
+				};
+				schema.items.properties.files =
+					this.files.map._descriptorPrototype_.type.prototype.toWSSchema();
```
kamsi: Files are NestedMaps if I recall correctly. For cases of NestedMap instances I think it is cleaner to aquire, the type info through `getItemType` method

model/form-entities-table.js
```diff
@@ -269,12 +269,31 @@ module.exports = memoize(function (db) {
 				isEntitiesNestedMap = entityObjects.value instanceof this.database.NestedMap;
 				if (!isEntitiesNestedMap) return result;
 				Item = entityObjects.value.getItemType();
+				// FIXME: This is probably broken. foreach loop over dataForms isn't correct.
+				// FIXME: it should be iteration over dataForm.map. Look under toWSSchema method.
```
kamsi: Good catch. Open a ticket for that, please. Even better if you can fix it in scope of that, but only if it won't take more thatn 30 minutes (so in case of any complications just leave it).

model/form-entities-table.js
```diff
@@ -269,12 +269,31 @@ module.exports = memoize(function (db) {
...
+					type: "array",
+					items: { type: "object", properties: {} }
+				};
+				entityObjects = this.propertyMaster.resolveSKeyPath(this.propertyName);
+				isEntitiesNestedMap = entityObjects.value instanceof this.database.NestedMap;
+				if (!isEntitiesNestedMap) return schema;
+				Item = entityObjects.value.getItemType();
+				Item.prototype.getBySKeyPath(this.sectionProperty).map.forEach(function (section) {
+					Object.assign(schema[this.propertyName].items.properties, section.toWSSchema());
+				}.bind(this));
```
kamsi: Just a side note, `forEach` will auto bind its second param, so you could write simply:

`...forEach(function (...) {..}, this)`

withoud explicit `blind`

model/form-section.js
```diff
@@ -313,6 +313,18 @@ module.exports = memoize(function (db) {
 
 				return result;
 			}
+		},
+		toWSSchema: {
+			value: function (ignore) {
+				var schema = {};
+				this.propertyNames.forEach(function (prop) {
```
kamsi: Maybe we should be explicit about operating on prototype in order to avoid any leaks from instance. So basically make sure that we operate on respective prototype in scope of sections.

model/document.js
raunokulla: @roxanadina Please specify how this should be filled. From [your comment](https://unctad.atlassian.net/browse/EV2CORE-113?focusedCommentId=17006&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17006) I do not completely understand how this should be filled. 

I can put dataforms title and its properties there, but [currently](https://github.com/egovernment/eregistrations/pull/1780/files/04e110b7dfdea7ff5266a5a2befb54c9a06367f8#diff-e1390d0bdaa281d91c2f9bb546767447R220) they are being added to schema.items.properties.data property. Should I remove that?

model/form-entities-table.js
raunokulla: duly noted 👍

model/document.js
roxanadina: @raunokulla 
this property should allow consumers to split the properties of the interesting entities in dataForms, as per our view model. 
I update this gist (manually..) with example
https://gist.github.com/roxanadina/c7b158a0336189845340ad667f7e7911#file-schema-els-json-L270-L279

while in nested entities, you should provide this prop at that specific level definition
https://gist.github.com/roxanadina/c7b158a0336189845340ad667f7e7911#file-schema-els-json-L315-L318

It should be more or less the same thing as we have in our view model, but respect the structure and properties of the schema.
These properties are right now only for project @mtuchowski is working on, so feel free to also ask him any questions

raunokulla: @roxanadina thanks for the explanation. Its clear now.

raunokulla:
Exported BusinessProcessCompany, BusinessProcessMerchant and BusinessProcessSolvency with current solution.
https://gist.github.com/raunokulla/783a77ed18c255e18b62df03527b3ece

Issue that caught my eye was that some dataForms do not have title property specified. For example [look dataForms property](https://gist.github.com/raunokulla/783a77ed18c255e18b62df03527b3ece#file-schemaexport-json-L2073) under "assistants" 

I traced the issue and found out that the cause of this is that nested entity Assistant dataForms map object 'assistants' [does not have label specified. ](https://github.com/egovernment/eregistrations-salvador/blob/master/model/business-process/nested-entities/assistant/data-forms/assistant.js#L19)

@roxanadina , @mtuchowski what would be correct behaviour in this situation?

raunokulla:
Discussed the solution with @mtuchowski and we realised that the solution currently is not handling DBJS objects as it should. Adding DBJS object schema export by Monday.

raunokulla:
Needs deepMerge module dependancy.

kamsi:
@roxanadina @mtuchowski I would like you to take a look at that, if all is good we need to incorporate `assignDeep` and then merge this one....

raunokulla:
https://gist.github.com/raunokulla/abf90fbb9bb62afeb34a01098b88a503
exported ELS schema.

roxanadina:
@raunokulla @mtuchowski @kamsi 
https://gist.github.com/raunokulla/abf90fbb9bb62afeb34a01098b88a503#file-gistfile1-txt-L873

we need a way to pass the system specific catalogs to this method

related to these pr also:
https://github.com/egovernment/eregistrations-salvador/pull/1882
https://github.com/egovernment/eregistrations-guatemala/pull/1126

kamsi:
@rauno @mtuchowski @roxanadina I don't think that we want to use `deepAssign` after all.
I looked here https://github.com/sindresorhus/deep-assign ... apart from quite some bad advert in the beginning it also doesn't look like it is maintained (last commit Jul 5, 2016). @raunokulla could you do another research?

PR #1821, opened 2017-04-19
Add "Date: ..." in first cell of the table in the page Files / Pending files

ilja903:


view/statistics-files-pending.js
```diff
@@ -66,7 +66,7 @@ exports['statistics-main'] = function () {
 	});
 	table({ class: 'statistics-table' },
 		thead(tr(
-			th(),
+			th(_('Date:'), ' ', window.moment().format('DD/MM/YYYY')),
```
kamsi: Please use what we usually use: `new db.Date()`. This way you should get the formatting right for every end system.

PR #1822, opened 2017-04-19
The rejection reasons disappears when I go to an other page

kamsi:
reported here: https://unctad.atlassian.net/browse/EV2CORE-7

The content of the rejection reasons disappears when I go to an other page and come back. To reproduce this bug, do the following: 1. go to Rejections reason page, 2. select "Since the start of the system" (dates are set from Sept 1st, 2016), 3. see the data (the table is filled), 4. go to an other page, 5. come back to Reasons of rejections → the period is the same, the table is empty.

### DIFF
--- a/view/components/filter-bar/select-date-range.js
+++ b/view/components/filter-bar/select-date-range.js
@@ -14,6 +14,7 @@
 var _ = require('mano').i18n.bind('Daterange')
   , dateFrom = require('./select-date-from')
   , dateTo = require('./select-date-to')
+  , location = require('mano/lib/client/location')
   , moment = window.moment
   , jQuery = window.jQuery;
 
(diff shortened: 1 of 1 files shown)


view/components/filter-bar/select-date-range.js
```diff
@@ -118,6 +119,20 @@ module.exports = function (/* opts */) {
 			$.dispatchEvent($('startId'), 'change');
 			$.dispatchEvent($('endId'), 'change');
 		});
+
+		var path;
+		path = location.pathname;
```
ilja903: why just not var path = location.pathname ? Maybe I do not see something

view/components/filter-bar/select-date-range.js
```diff
@@ -118,6 +119,20 @@ module.exports = function (/* opts */) {
...
+		location.on('change', function () {
+			if (location.pathname !== path) {
+				path = location.pathname;
```
ilja903: if path from the top is always location.pathname then 
location.pathname !== path ==> location.pathname !== location.pathname does not make any sence ie it is always false

ilja903:
Everything else seems fine

view/components/filter-bar/select-date-range.js
kamsi: Yeah weird habit.

PR #1828, opened 2017-04-20
Remove the "Withdraw" role div in Times / per role and in Times / per person

ilja903:
https://unctad.atlassian.net/browse/EV2CORE-32


view/statistics-time-per-person.js
```diff
@@ -39,14 +39,25 @@ var getRowResult = function (rowData, label) {
 	return result;
 };
 
+var processingStepsMetaFrontDeskFilter = function (processingStepsMeta) {
```
kamsi: Yeah, OK I think I ill advised you last time. Let's make a util out of this. PLace it in `view/utils`

view/statistics-time-per-person.js
```diff
@@ -39,14 +39,25 @@ var getRowResult = function (rowData, label) {
...
+	var processingStepsMetaWithoutFrontDesk = {};
+	Object.keys(processingStepsMeta).forEach(function (key) {
+		if (key !== 'frontDesk') {
```
kamsi: use this module `require('es5-ext/string/#/starts-with')` to check for `frontDesk`... there are steps which are treated as frontDesk and have `frontDesk` only as prefix... we want to filter them as well.

kamsi:
@ilja903 

```javascript

'use strict';
var processingStepsMeta = require('../../processing-steps-meta');
var startsWith = require('es5-ext/string/#/starts-with');

module.exports = function () {
	var processingStepsMetaWithoutFrontDesk = {};
	Object.keys(processingStepsMeta).forEach(function (key) {
		if (!startsWith.call(key, 'frontDesk')) {
			processingStepsMetaWithoutFrontDesk[key] = processingStepsMeta[key];
		}
	});
	return processingStepsMetaWithoutFrontDesk;
};

```

This should do the trick. then you just require it and invoke to get the steps.

PR #1832, opened 2017-04-21
Too much space below the filters

ilja903:
https://unctad.atlassian.net/browse/EV2CORE-38


ilja903:
if you see fast track you can merge directly and put the task into ready to jira.

PR #1831, opened 2017-04-21
Prepare new data representation to be sent for Times / Per role and Times / Roles - operator tables

mtuchowski:
In both "Times / Per role" and in "Times / Roles - operator" tables, something would be very useful for us is to be able to see, for each "Processed file", what was the file, it operator, it time of processing, the start time and the end time based on which are calculated the time.

Proposals for display for Roles table : http://3m5fo4.axshare.com/#p=in_roles, for Roles/operators table: http://3m5fo4.axshare.com/#p=in_roles___operators (use as data specification)

### DIFF
--- a/server/business-process-query/get-data.js
+++ b/server/business-process-query/get-data.js
@@ -256,6 +256,12 @@ exports.stepMetaMap = {
 };
 
 exports.businessProcessMetaMap = {
+	businessName: {
+		type: 'computed',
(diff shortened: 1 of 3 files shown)


mtuchowski:
@kamsi Wait a bit, I assigned you too fast.

mtuchowski:
@kamsi OK, added missing bits.

PR #1833, opened 2017-04-21
Enable mongo-db in the systems

kamsi:


kamsi:
@roxanadina @mtuchowski probably we can merge this (will not do much harm), but we still need to put mongo in node_modules of all the systems....

kamsi:
@mtuchowski 

1. I don't see such tool
2. We will always be able to move our data to a new storage (and the migration should not be too painful as long as we choose sth which stores JSON).

kamsi:
@mtuchowski I don't see need for mongo setup in ldz. Still I will update it's dependency, as it is our point of reference.

kamsi:
@mtuchowski adjusts ready.

kamsi:
@mtuchowski can you merge this one?

PR #1835, opened 2017-04-24
Hide the "Pending" option in the selector, and have the "Validated" option selected by default.

ilja903:
https://unctad.atlassian.net/browse/EV2CORE-41


view/statistics-flow-roles.js
```diff
@@ -141,6 +141,9 @@ exports['statistics-main'] = function () {
 						select(
 							{ id: 'step-status', name: 'status' },
 							list(Object.keys(stepStatuses), function (status) {
+								if (status === 'pending') {
```
kamsi: I'm can't say I'm overjoyed with this solution, as you could simply remove a string from config in line 59.

ilja903:
You are totally right, that was silly

PR #1836, opened 2017-04-24
Remove the hours in the dates of the reasons of rejection.

ilja903:
https://unctad.atlassian.net/browse/EV2CORE-5

### DIFF
--- a/view/utils/last-modified.js
+++ b/view/utils/last-modified.js
@@ -4,5 +4,5 @@ var db = require('mano').db;
 
 module.exports = function (date) {
 	if (!date) return;
-	return String(new db.DateTime(Math.round(date / 1000)));
+	return String(new db.Date(Math.round(date / 1000)));
 };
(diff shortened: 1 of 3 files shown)


view/utils/last-modified.js
```diff
@@ -4,5 +4,5 @@ var db = require('mano').db;
 
 module.exports = function (date) {
 	if (!date) return;
-	return String(new db.DateTime(Math.round(date / 1000)));
+	return String(new db.Date(Math.round(date / 1000)));
```
kamsi: @ilja903 this change has quite far going consequences as it changes what is displayed other places (I will only point ones from eregistrations, but there may be others in scope of end systems):

https://github.com/egovernment/eregistrations/blob/master/view/components/table-columns.js#L120

https://github.com/egovernment/eregistrations/blob/master/view/manager-business-processes.js#L151

I believe the proposed change is to wide...

However in order to maintain control over this util and reuse it, you could add options to it.
By default the util would have worked as it did before but when called with option `useDate` or `dateOnly` or maybe a better name, then it would produce result from `Date` instead of `DateTime`.

PR #1830, opened 2017-04-24
New view for rejection reasons

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-43

We need to prepare the view to handle custom JSON from the server instead of what we have for that table now.


view/statistics-time-per-person.js
```diff
@@ -115,36 +115,37 @@ exports['statistics-main'] = function () {
...
-							td({ class: 'statistics-table-number' }, rowData.avgTime),
-							td({ class: 'statistics-table-number' }, rowData.minTime),
-							td({ class: 'statistics-table-number' }, rowData.maxTime)
-						);
-					})
-					));
-		});
-	}));
+				))),
+		br(),
```
mtuchowski: It was agreed some time ago that we should not use hard line breaks in views and use css for that.

server/routes/statistics.js
```diff
@@ -242,36 +237,16 @@ module.exports = function (config) {
...
-
-						return result;
-					});
-				});
+				var data = { rows: [], pageCount: 1 };
+				data.pageCount = getPage(data.rows, query.page);
+				/* TODO inflate data.rows with real results
+					expected structure:
+					[
+					['Reason 1, Reason 2', '*', 3, 'John Smith', 'Revision', '02/03/2017', 'Super inc.'],
```
mtuchowski: Let's maybe use hashes instead of arrays of values that don't transfer any information about the meaning of each value.

view/statistics-flow-rejections.js
```diff
@@ -1,12 +1,17 @@
 'use strict';
 
 var _                            = require('mano').i18n.bind('View: Statistics')
-  , env                          = require('mano').env
-  , location                     = require('mano/lib/client/location')
-  , selectService                = require('./components/filter-bar/select-service')
-  , tableColumns                 = require('./components/statistics-rejections-table-columns')
```
mtuchowski: If we would use the hashes instead of array's of values then we will probably need some simple table columns definition for display. Or a mapper of some sorts.

server/routes/statistics.js
kamsi: @mtuchowski that was exactly what I wanted to achieve: no manipulation of data on the view, jsut display. This way we force controller to give us all we need, and all the views that will follow (pdf, csv) will have an easier task with parsing. So I would rather leave it as is.

kamsi:
@mtuchowski prepared separate ticket for brs https://github.com/egovernment/eregistrations/issues/1847

PR #1837, opened 2017-04-24
Date selector throwing errors

ilja903:
https://unctad.atlassian.net/browse/EV2CORE-52


ilja903:
I do not know why page without period selector fires setTimeout.  Maybe you have better idea?

ilja903:
It happened on a page where you review files (Revision of the documents tab) When you click on a table row with file. Each time you click on a different row the error occures

PR #1818, opened 2017-04-25
Create service to save rejection reasons history for uploads

raunokulla:
This task is about creating a service to track history of rejection reasons for a requirement upload.

Right now we simply store rejection reasons as a multiple value for [requirement uploads](https://github.com/egovernment/eregistrations/blob/master/model/requirement-upload.js#L70)
To achieve what is requested in this ticket we will need to keep track of history of changes of this `rejectReasons` property.
To achieve this we should create a service which stores this history.
Please see [this](https://github.com/egovernment/eregistrations/blob/master/server/services/status-history-logger.js) for an inspiration on how such service may be configured:


raunokulla:
output:
```
{
  "hasOnlySystemicReasons": false,
  "rejectionType": "rejected",
  "rejectionReasons": [
    {
      "types": [
        "other"
      ],
      "value": "everything is not good.",
      "ownerType": "processingStep"
    },
    {
      "types": [
        "other"
      ],
      "value": "data is not good.",
      "ownerType": "data"
    },
    {
      "types": [
        "illegible"
      ],
      "value": "",
      "ownerType": "requirementUpload"
    },
    {
      "types": [
        "other"
      ],
      "value": "not good",
      "ownerType": "requirementUpload"
    },
    {
      "types": [
        "other"
      ],
      "value": "not sufficient",
      "ownerType": "paymentReceipt"
    }
  ],
  "service": {
    "type": "BusinessProcessCompany",
    "id": "51bmuii98iz4",
    "businessName": "2312312"
  },
  "operator": {
    "id": "218nvxux6gor",
    "name": "Rauno Kulla"
  },
  "processingStep": {
    "path": "/processingSteps/map/revision",
    "label": "Revision"
  },
  "date": 1493155310567
}
```

server/services/query-memory-db/queries/business-process-rejection-reasons.js
```diff
@@ -0,0 +1,110 @@
...
+
+			}
+			if (processingStep.isRejected) {
+				result.hasOnlySystemicReasons = false;
+				result.rejectionReasons.push(rejectionReason('processingStep', ['other'], otherValue));
+			}
+			return true;
+		});
+		//rejectionReasons -> data
+		if (businessProcess.dataForms.rejectReason) {
```
kamsi: This should be redundant. If bp was rejected it must have a `isRejected === true` on some step. And that case is already taken care of in the loop over steps. Furthermore, this will duplicate the entry to `result.rejectionReasons`.

server/services/query-memory-db/queries/business-process-rejection-reasons.js
```diff
@@ -0,0 +1,110 @@
...
+		//rejectionReasons -> paymentReceiptUploads
+		businessProcess.paymentReceiptUploads.applicable.forEach(function (paymentReceiptUpload) {
+			if (paymentReceiptUpload.status !== 'invalid') return;
+			result.hasOnlySystemicReasons = false;
+			result.rejectionReasons.push(
+				rejectionReason('paymentReceipt', ['other'], paymentReceiptUpload.rejectReasonMemo)
+			);
+		});
+
+		result.date = Date.now();
```
kamsi: Use `new db.Date()`... make sure that you write down proper ts in milliseconds. Basically what you save here is server time. We use UTC everywhere for saving data (db.Date ensures that + it is what we normally use).

server/services/status-history-logger.js
```diff
@@ -41,6 +42,23 @@ var storeLog = function (storage, logPath/*, options */) {
...
+	if (event.type !== 'direct') return;
+	if (!event.path.startsWith('processingSteps')) return;
+	status = unserializeValue(event.data.value);
+	if (status !== 'rejected' && status !== 'sentBack') return;
+
+	return queryMemoryDb([event.ownerId], 'businessProcessRejectionReasons', {
+		businessProcessId: event.ownerId
+	}).done(function (reasonObject) {
+		//TODO here should be added logic for saving the reasonObject
+		console.log(JSON.stringify(reasonObject));
```
kamsi: OK, mongo is already there so let's use it! https://github.com/egovernment/eregistrations/pull/1833

server/services/query-memory-db/queries/business-process-rejection-reasons.js
kamsi: As discussed, this is actually needed, ignore my previous comment.

server/services/query-memory-db/queries/business-process-rejection-reasons.js
```diff
@@ -0,0 +1,109 @@
...
+		},
+		processingStep: {
+			path: null,
+			label: null
+		},
+		date: null
+	};
+
+	var setProcessingStepData = function (step) {
+		result.processingStep.label = step.label;
```
kamsi: Let's pass result directly to function, so there is no doubt as to what we're talking about here

server/services/query-memory-db/queries/business-process-rejection-reasons.js
```diff
@@ -0,0 +1,113 @@
...
+		businessProcess.paymentReceiptUploads.applicable.forEach(function (paymentReceiptUpload) {
+			if (paymentReceiptUpload.status !== 'invalid') return;
+			result.hasOnlySystemicReasons = false;
+			result.rejectionReasons.push(
+				rejectionReason('paymentReceipt', ['other'], paymentReceiptUpload.rejectReasonMemo
+				  , paymentReceiptUpload.__id__.slice(paymentReceiptUpload.__id__.indexOf('/')))
+			);
+		});
+
+		result.date = Number(new db.DateTime());
```
kamsi: `Number(new db.Date())`

we don't want whole `DateTime` - we want date without time on the view, and we want to group by it later on.

server/services/status-history-logger.js
```diff
@@ -41,10 +43,29 @@ var storeLog = function (storage, logPath/*, options */) {
...
 };
 
+var saveRejectionReason = function (event) {
+	var status;
+	if (event.type !== 'direct') return;
+	if (!event.path.startsWith('processingSteps')) return;
+	status = unserializeValue(event.data.value);
+	if (status !== 'rejected' && status !== 'sentBack') return;
+
+	return mano.queryMemoryDb([event.ownerId], 'businessProcessRejectionReasons', {
```
kamsi: Let's require queryMemoryDb up front like we usally do in the beginning of the module. Also, no need to return anything as we consume with `done` (what you return here right now is `undefined` [as that is what is returned from done]).

PR #1850, opened 2017-04-28
Fix mongo-db missing env

mtuchowski:


kamsi:
@mtuchowski great job finding it fast, thank you.

PR #1849, opened 2017-04-28
Count occurences in the rejection reasons

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-2

We will need to prepare a script which calculates it and saves it. The script should operate only in scope of mongo, can be launched at start of the server.

We want to group by rejection reasons (data from first row) and by date.


server/services/recalculate-rejection-reasons-occurrences.js
```diff
@@ -0,0 +1,51 @@
...
+					}
+				}
+			]).toArray();
+		}).then(function (result) {
+			return deferred.map(result, function (item) {
+				if (item.count <= 1) return;
+				return collection.update({ 'date.date': item._id.date,
+					rejectionReasonsConcat: item._id.rejectionReasonsConcat
+					},
+					{ $set: { occurrencesCount: item.count } });
```
roxanadina: and @kamsi third param: {multi:true}
https://docs.mongodb.com/manual/reference/method/db.collection.update/#multi-parameter

server/services/recalculate-rejection-reasons-occurrences.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+var mongo    = require('../mongo-db')
+  , Set      = require('es6-set')
+  , deferred = require('deferred');
+
+module.exports = function () {
```
mtuchowski: If I understand correctly this is supposed to be a one time migration right?

kamsi: @mtuchowski no it's not. It is supposed to update `rejectionReasonsConcat` on every server start in order to facilitate grouping query search.

Check out the adjusts...

mtuchowski:
OK, waiting for adjusts to be approved.

PR #1843, opened 2017-04-28
Rejection reasons - attach view to data

raunokulla:
After that step we should have the new rejection reasons table fully working with the gathered data.


server/routes/statistics.js
```diff
@@ -236,17 +238,22 @@ module.exports = function (config) {
...
+			}).then(function (count) {
+				var offset, limit;
+				offset = (page - 1) * itemsPerPage;
+				if (page * itemsPerPage < count) limit = itemsPerPage;
+				data.pageCount = ensureNumber(Math.ceil(count / itemsPerPage));
+				return getRejectionReasons.find(queryData, offset, limit);
+			}).then(function (reasons) {
+				data.rows = parseRejectionsForView(reasons);
 				return data;
+			}, function (err) {
```
kamsi: Actually this error callback should not be needed. Like in case of other controller handlers we should return a promise or the final data... the controller generic mechanisms should catch errors.

server/services/query-memory-db/queries/business-process-rejection-reasons.js
```diff
@@ -38,7 +38,7 @@ module.exports = exports = function (db) {
 				path: null,
 				label: null
 			},
-			occurancesCount: 0,
+			occurrancesCount: 0,
```
kamsi: not `occurrancesCount`, but `occurrencesCount`

utils/statistics-flow-rejection-reason-results.js
```diff
@@ -0,0 +1,25 @@
...
+			reasonItem.types.forEach(function (type) {
+				reasonsConcat.push(type);
+				if (type === 'other' && reasonItem.value) {
+					reasonsConcat.push(reasonItem.value);
+				}
+			});
+		});
+		result.push(reasonsConcat);
+		result.push(reason.hasOnlySystemicReasons ? '*' : '');
+		result.push('');
```
kamsi: I don't expect much data here, but let's be prepared just in case :)... so basically I think we should show sth here when it is a number greater than 1.

utils/statistics-flow-rejection-reason-results.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+var db = require('../db');
+
+module.exports = function (reasons) {
+	return reasons.map(function (reason) {
+		var result = [], reasonsConcat = [];
+		reason.rejectionReasons.forEach(function (reasonItem) {
```
kamsi: I think that what was also expected was to include some kind information whether given reason was about data or document: https://unctad.atlassian.net/browse/EV2CORE-3

`Document 1 - the document is not readable", "Document N - the document does not correspond to the expected document", etc. "Data - lorem ipsum lorem ipsum, etc."`

Let's add Document N - and Data - suffixes, and nothing for the main reason if we have one.

utils/statistics-flow-rejection-reason-results.js
raunokulla: Please explain what are you thinking of.

utils/statistics-flow-rejection-reason-results.js
```diff
@@ -0,0 +1,35 @@
...
+					return;
+				}
+				reasonsConcat.push(
+					prefix + db.RequirementUploadRejectReason.meta[type].label
+				);
+			});
+		});
+		result.push(reasonsConcat);
+		result.push(reason.hasOnlySystemicReasons ? '*' : '');
+		result.push(reason.occurrencesCount > 0 ? reason.occurrencesCount : '');
```
kamsi: reason.occurrencesCount > 1

PR #1851, opened 2017-05-02
Fix filter that goes to next line in Files/Roles

ilja903:
https://unctad.atlassian.net/browse/EV2CORE-60


ilja903:
idea is the same as in https://github.com/egovernment/eregistrations/pull/1834/files

PR #1854, opened 2017-05-04
Adjustments after tests

kamsi:
As described in the comment here: 

https://unctad.atlassian.net/browse/EV2CORE-3

1. we have lost the design of the table on your test port
2. the table is placed in a section that should be of class table-responsive-container (not doing this is going backwards for no reason) and not section-primary
3. we have lost the last column with the magnifying glass from which we could open the file

### DIFF
--- a/view/statistics-flow-rejections.js
+++ b/view/statistics-flow-rejections.js
@@ -80,21 +80,21 @@ exports['statistics-main'] = function () {
 			section({ class: 'pad-if-pagination' }, pagination),
 			br(),
 			data.map(function (result) {
+				var defaultOpts = { class: "submitted-user-data-table-date-time" };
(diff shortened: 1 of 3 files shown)


view/statistics-flow-rejections.js
```diff
@@ -80,30 +80,40 @@ exports['statistics-main'] = function () {
...
+					table({ class: 'submitted-user-data-table' },
 						thead(
-							th(_("Rejection reasons")),
-							th(),
-							th(),
-							th(_("Operator")),
-							th(_("Role")),
-							th(_("Date")),
-							th(_("File"))
+							th({ class: "submitted-user-data-table-date-time" }, _("Rejection reason")),
```
raunokulla: You should use defaultOpts for headers as well.

PR #1859, opened 2017-05-04
rejection reasons statistics csv export formation issues

raunokulla:
When the explanation of rejection reason contains single or double quotes in it or when last column (business name) contains single or double quotes or comma, it will break the CSV formation.

These fields have to be surrounded by double quotes and all existing double quotes in them have to be replaced with single quotes. Then it also no longer matters whether these columns contain commas. Source: http://stackoverflow.com/a/8497474

### DIFF
--- a/server/csv-renderers/statistics-flow-rejections.js
+++ b/server/csv-renderers/statistics-flow-rejections.js
@@ -14,8 +14,10 @@ module.exports = function (result, config) {
 	];
 
 	data.push.apply(data, result.map(function (row) {
-		row[0] = '\"' + row[0].join(' ').replace(/\r?\n|\r/g, ' ').replace(/"/g, "'") + '\"';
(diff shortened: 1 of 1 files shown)


server/csv-renderers/statistics-flow-rejections.js
```diff
@@ -14,7 +14,8 @@ module.exports = function (result, config) {
 	];
 
 	data.push.apply(data, result.map(function (row) {
-		row[0] = row[0].join(' ').replace(/\r?\n|\r/g, ' ').replace(/,/g, ' ');
+		row[0] = '\"' + row[0].join(' ').replace(/\r?\n|\r/g, ' ').replace(/"/g, "'") + '\"';
```
kamsi: In order to make it future proof (maybe some other cols added, or some suspicious sign in other cells) and clean let's use sanitation performed on this row, for all the rows.

PR #1862, opened 2017-05-07
Enable options support in get-soap-client utils.

raunokulla:
the util that handles creating soap client has to be modified in a way that it would support passing options argument to soap module's createClient() function.
By doing so it will allow us to adjust the default soap client generation as we deem necessary.

First use case occurred in https://github.com/egovernment/eregistrations-salvador/issues/1808 where it's necessary to force namespace prefix usage in methodName in order to receive proper results from the service.

JIRA issue: https://unctad.atlassian.net/browse/ELS-25

### DIFF
--- a/utils/get-soap-client.js
+++ b/utils/get-soap-client.js
@@ -11,10 +11,10 @@ var memoize          = require('memoizee')
 
   , createSoapClient = promisify(soap.createClient);
 
-var getSoapClientImpl = memoize(function (wsdlUrl, soapUrl) {
-	debug('creating soap client (%s)', soapUrl);
+var getSoapClientImpl = memoize(function (wsdlUrl, options) {
+	debug('creating soap client (%s)', options.endpoint);
 
-	return createSoapClient(wsdlUrl, { endpoint: soapUrl })(function (client) {
+	return createSoapClient(wsdlUrl, options)(function (client) {
 		var services = client.wsdl.definitions.services
 		  , ports, methods, method, promisifiedMethodName;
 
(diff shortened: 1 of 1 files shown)


utils/get-soap-client.js
```diff
@@ -11,10 +11,10 @@ var memoize          = require('memoizee')
 
   , createSoapClient = promisify(soap.createClient);
 
-var getSoapClientImpl = memoize(function (wsdlUrl, soapUrl) {
-	debug('creating soap client (%s)', soapUrl);
+var getSoapClientImpl = memoize(function (wsdlUrl, options) {
```
kamsi: As the name suggests, `options` should be optional, we don't expose them in the signature of functions, other than in comment.

utils/get-soap-client.js
```diff
@@ -11,10 +11,10 @@ var memoize          = require('memoizee')
...
+	debug('creating soap client (%s)', options.endpoint);
 
-	return createSoapClient(wsdlUrl, { endpoint: soapUrl })(function (client) {
+	return createSoapClient(wsdlUrl, options)(function (client) {
```
kamsi: same here

kamsi:
@raunokulla I cannot see reference to Jira, so please provide one - also change status in Jira if needed.

raunokulla:
JIRA reference added.
Also since this modification does not affect any current end system. All functionalities that used the get-soap-client util will be able to use it in the same way.
This functionality is needed for  egovernment/eregistrations-salvador#1808 and this ticket is already referenced there.

PR #1866, opened 2017-05-11
Enable querying for pending rows

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-80

### DIFF
--- a/apps-common/query-conf/date-from.js
+++ b/apps-common/query-conf/date-from.js
@@ -13,11 +13,10 @@ module.exports = {
 		}
 		if (!value) {
 			if (dateTo) {
-				return new db.Date(dateTo.getUTCFullYear(), dateTo.getUTCMonth(),
-						dateTo.getUTCDate() - 6);
+				return new db.Date(dateTo.getUTCFullYear());
 			}
-			// last week by default
-			return new db.Date(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate() - 6);
+			// current year by default
+			return new db.Date(now.getUTCFullYear());
 		}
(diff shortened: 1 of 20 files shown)


apps-common/query-conf/date-from.js
```diff
@@ -13,11 +13,10 @@ module.exports = {
...
 		if (!value) {
 			if (dateTo) {
-				return new db.Date(dateTo.getUTCFullYear(), dateTo.getUTCMonth(),
-						dateTo.getUTCDate() - 6);
+				return new db.Date(dateTo.getUTCFullYear());
 			}
-			// last week by default
-			return new db.Date(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate() - 6);
+			// current year by default
+			return new db.Date(now.getUTCFullYear());
```
mtuchowski: This will set the date close to timestamp 0 (1/1/1970). Consider the following:

```javascript
> var d = new db.Date()
> console.log(new db.Date(d.getUTCFullYear()));
1/1/1970
```

This is because in single argument mode db.Date treats it as timestamp.

utils/to-date-time-in-time-zone.js
```diff
@@ -6,7 +6,7 @@
 var ensureDate   = require('es5-ext/date/valid-date')
   , ensureString = require('es5-ext/object/validate-stringifiable-value');
 
-module.exports = function (date, timeZone) {
+module.exports = function (date, timeZone, debug) {
```
mtuchowski: A leftover?

kamsi:
@mtuchowski fix applied - good catch. The functionality passed the tests, so you can merge at will.

PR #1869, opened 2017-05-15
Skip part-A step if applicable document size equals 0

raunokulla:
In some of the end-systems it is possible that the size of applicable requirementUploads list is 0 in scope of one business process. In this situation user should not be directed to document upload step after he/she has finished filling forms.

It is also necessary to account with situation that this business process does not contain any costs. In this case payment step has to be skipped as well. 

The necessity of this modification came from: https://unctad.atlassian.net/browse/ELS-53

### DIFF
--- a/view/business-process-data-forms.js
+++ b/view/business-process-data-forms.js
@@ -14,7 +14,9 @@ exports._parent = require('./business-process-base');
 exports.step = function () {
 	var businessProcess = this.businessProcess
 	  , dataForms       = businessProcess.dataForms
-	  , guideProgress   = businessProcess._guideProgress;
+	  , guideProgress   = businessProcess._guideProgress
+	  , applicableDocs  = businessProcess.requirementUploads.applicable._size;
+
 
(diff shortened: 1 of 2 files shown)


view/business-process-data-forms.js
```diff
@@ -31,7 +32,13 @@ exports.step = function () {
 	insert(_if(and(eq(guideProgress, 1),
 		eq(dataForms._progress, 1)),
 		div({ class: 'user-next-step-button' },
-			a({ href: '/documents/' }, _("Continue to next step"))),
+			_if(eq(applicableDocs, 0),
+				_if(eq(businessProcess.costs._paymentWeight, 0),
+					a({ href: '/submission/' }, _("Continue to next step")),
```
mtuchowski: You can just wrap the href value in `_if`s. Like:

```javascript
a({ href: _if(…) }, _("Continue to next step")))
```

This will limit the copy paste.

raunokulla: Modified accordingly.

PR #1870, opened 2017-05-17
Make the details look different than the other rows

ilja903:
https://unctad.atlassian.net/browse/EV2CORE-117
- have zero margin, so it does not look like a "table into a table"
- a grey background to make it pop out differently than the normal lines

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -274,9 +274,3 @@ script { display: none !important; }
 .overflow-x{
 	overflow-x: auto;
 }
-
-/*csslint important: false*/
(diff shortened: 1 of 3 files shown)


css/base.css
```diff
@@ -274,3 +274,9 @@ script { display: none !important; }
 .overflow-x{
 	overflow-x: auto;
 }
+
+/*csslint important: false*/
+.no-padding{
+	padding: 0 !important;
```
kamsi: Provide a hackless solution, please.

PR #1875, opened 2017-05-18
Fix period group selector

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-75


kamsi:
@mtuchowski believe me, I tried different options - they didn't work.

PR #1876, opened 2017-05-18
Server crash when in legacy in stats

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-121


view/components/filter-bar/select-date-range.js
```diff
@@ -40,6 +40,7 @@ module.exports = function (/* opts */) {
 			console.error('Probably you will have to integrate JQuery and all ' +
 				'rangepicker dependencies into into index.html.tpl file. ' +
 				'Check select-date-range.js for instructions.');
+			return;
```
mtuchowski: Oooo…

PR #1883, opened 2017-05-24
Duplicate the table in the Dashboard to show the details per certificate

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-124

mockup: http://3m5fo4.axshare.com/#p=dashboard (second table)

### DIFF
--- a/server/business-process-query/get-status-history-date-map.js
+++ b/server/business-process-query/get-status-history-date-map.js
@@ -44,7 +44,7 @@ Date map layout:
 		withdrawn: sum,
 		// 2.6 On business process - At rejected
 		rejected: sum,
-		// Notice: Not on business process, but on actual certificate approved
+		// 2.7 Notice: Not on business process, but on actual certificate approved
 		approved: sum
 	},
 	// Processing step stauses: pending, paused, sentBack, rejected, approved, redelegated
@@ -225,6 +225,7 @@ var getDateMap = function (data, statusHistoryData) {
 				var logStauts = statusHistoryLog.status
 				  , logDate   = statusHistoryLog.date;
 
+				// 2.7 [date][serviceName].certificate[certificateName].approved
 				if (logStauts === 'approved') {
 					getCertificateDataset(logDate)[logStauts]++;
 				}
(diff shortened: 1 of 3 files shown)


server/business-process-query/get-status-history-date-map.js
```diff
@@ -43,7 +43,9 @@ Date map layout:
 		// 2.5 On business process - At withdrawn or closed
 		withdrawn: sum,
 		// 2.6 On business process - At rejected
-		rejected: sum
+		rejected: sum,
+		// Notice: Not on business process, but on actual certificate approved
+		approved: sum
```
mtuchowski: It would mean a great deal to me if you could keep using that numbering I introduced for easier debugging.

server/business-process-query/get-status-history-date-map.js
```diff
@@ -223,6 +225,10 @@ var getDateMap = function (data, statusHistoryData) {
 				var logStauts = statusHistoryLog.status
 				  , logDate   = statusHistoryLog.date;
 
+				if (logStauts === 'approved') {
+					getCertificateDataset(logDate)[logStauts]++;
```
mtuchowski: So a comment here with probably `2.7` in it.

server/routes/statistics.js
```diff
@@ -120,6 +121,20 @@ var businessProcessQueryHandler = new QueryHandler([{
 	}
 }]);
 
+var getPeriods = function () {
+	var periods = ['inPeriod', 'today', 'thisWeek', 'thisMonth']
+	  , today = toDateInTz(new Date(), db.timeZone)
+	  , currentYear = new db.Date(today.getUTCFullYear(), 0, 1)
+	  , lastYearInRange = new db.Date(today.getUTCFullYear() - 5, 0, 1);
```
mtuchowski: It will be outdated in less than a year. I think it's better to hardcode that one in. This would mean changing `view/statistics-dashboard.js` as well. I understand that we want to show all years from beginning of the system, not only last 5 years. Maybe ask Vianney about it.

kamsi: I asked, and we established that last 6 years (including current) is totally fine solution. Otherwise with each year the table becomes less and less readable.

PR #1884, opened 2017-05-26
Some rejections have no reason given

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-134


kamsi:
@mtuchowski  should be taken together with: https://github.com/egovernment/eregistrations-salvador/issues/1867

PR #1887, opened 2017-05-29
fix toMetaDataJSON method for form-entities-table

raunokulla:
Inside that method iteration over formSections is not being done correctly.
It is done just over dataForms but instead it should be done over dataForms.map


raunokulla:
Tested it with previous code. Wasn't working. It said that section does not have toMetaDataJSON function as we excpected.
Tested it with new approach. It returns correctly expected output.

PR #1891, opened 2017-05-30
Add sorting feature to "Files / Certificates"

ilja903:
https://unctad.atlassian.net/browse/EV2CORE-106


view/utils/tablesorter-time-range-comarator.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
+
+function contains(a, searchString) {
```
kamsi: Basically this whole module looks like a lot of code do what is needed.

I propose something quite simpler:

```javascript
var getTime = function (res) {
	var time = 0; // in seconds
	if (res.indexOf('<') !== -1) {
		return -1;
	}
	if (res.indexOf('h') !== -1) {
		res = res.split('h');
		time += Number(res[0]) * 3600;
		res = res[1];
	}
	if (res.indexOf('m') !== -1) {
		time += Number(res.split('m')[0]) * 60;
	}

	return time;
};

module.exports = function (a, b) {
	var resA, resB, timeA = 0, timeB = 0;
	resA = a.replace(/\s+/g, '');
	resB = b.replace(/\s+/g, '');

	timeA = getTime(resA);
	timeB = getTime(resB);
	
	return timeA - timeB;
};
```

This sorts ascending....

kamsi: Also tests would be welcome... for example:

[ "6 h 35 m", "6 h 34 m", "34 m", "14 h 41 m", "20 h", "< 1m", "10 m", "1 h"]

should sort in the following way:

["< 1m", "10 m", "34 m", "1 h", "6 h 35 m", "6 h 34 m", "14 h 41 m", "20 h"]

view/utils/init-table-sorting-on-client.js
```diff
@@ -1,12 +1,9 @@
...
-	if (!window.jQuery) return;
-	var checkExist = setInterval(function () {
-		var element = window.jQuery(jQuerySelector);
-		if (element.length) {
-			element.tablesorter();
-			clearInterval(checkExist);
-		}
+module.exports = function (domElement, tableSorterOpts) {
+	setInterval(function () {
+		var element = window.jQuery(domElement);
```
kamsi: This one is still not addressed. We need sth like:
`if (!window.jQuery) return;`

before we even call `setInterval`

conf/objects-list-all-on-one-page.js
```diff
@@ -0,0 +1,3 @@
+'use strict';
```
kamsi: I don't see this file required anywhere, so it cannot affect the files / certificates table. We want a fully working functionality for files / certificates after this PR is merged.

conf/objects-list-unlimited-items-per-page.js
```diff
@@ -0,0 +1,3 @@
+'use strict';
+
+module.exports = 100000;
```
mtuchowski: This is a little strange choice for a value. Particularly because it's used to represent `unlimited` number of items. This imposes that it should probably be infinity. As we can't use that I suggest using maybe `Number.MAX_SAFE_INTEGER` or other constant that gets a little bit closer to that.

PR #1890, opened 2017-05-30
Add approved column to Files / Certificates

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-142


roxanadina:
@kamsi , I have less files when "All inscriptions" than when selecting Nit, for example.
is this suppose to be a cumulative condition (nb files that have all inscription approved vs a specific inscription )

kamsi:
@roxanadina you are correct. When we don't filter by inscription we consider the approved business processes... bp is approved when it has status 'closed' https://github.com/egovernment/eregistrations/blob/master/model/business-process-new/flow.js#L116

Once we apply cert filter, we calculate bps which got the specific cert approved: https://github.com/egovernment/eregistrations/blob/master/model/document.js#L121-L128 (so, when either bp is approved, or the processingStep of the certificate is approved).

PR #1894, opened 2017-06-01
Enable sorting for roles, operators and rejections

kamsi:
IT corresponds to following Jira tickets:

https://unctad.atlassian.net/browse/EV2CORE-107
https://unctad.atlassian.net/browse/EV2CORE-108
https://unctad.atlassian.net/browse/EV2CORE-109

### DIFF
--- a/view/utils/init-table-sorting-on-client.js
+++ b/view/utils/init-table-sorting-on-client.js
@@ -16,6 +16,23 @@ module.exports = function (domElement/*, opts */) {
 		// set type, either numeric or text 
 		type: 'numeric'
 	});
+	window.jQuery.tablesorter.addParser({
+		// set a unique id 
+		id: 'dates',
+		is: function (s) {
+			// return false so this parser is not auto detected 
+			return false;
+		},
+		format: function (item) {
+			var parsedItem = item;
+			if (item.indexOf('/') !== -1) {
+				parsedItem = item.replace(/\/+/g, '-').split('-').reverse().join('-');
+			}
+			return new Date(parsedItem);
+		},
+		// set type, either numeric or text 
+		type: 'numeric'
+	});
 	var opts = normalizeOptions(arguments[1]);
 	setInterval(function () {
 		var element = window.jQuery(domElement);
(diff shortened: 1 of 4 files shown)


view/utils/init-table-sorting-on-client.js
```diff
@@ -16,6 +16,23 @@ module.exports = function (domElement/*, opts */) {
 		// set type, either numeric or text 
 		type: 'numeric'
 	});
+	window.jQuery.tablesorter.addParser({
+		// set a unique id 
+		id: 'dates',
+		is: function (s) {
+			// return false so this parser is not auto detected
```
mtuchowski: While stating that returning false will prevent auto-detection is valuable information I would like to know why we don't want it to be auto-detected as well...

view/utils/init-table-sorting-on-client.js
```diff
@@ -16,6 +16,23 @@ module.exports = function (domElement/*, opts */) {
...
+		// set a unique id 
+		id: 'dates',
+		is: function (s) {
+			// return false so this parser is not auto detected 
+			return false;
+		},
+		format: function (item) {
+			var parsedItem = item;
+			if (item.indexOf('/') !== -1) {
+				parsedItem = item.replace(/\/+/g, '-').split('-').reverse().join('-');
```
mtuchowski: Don't we have any locale dependent formatter for dates in JS/our system?

mtuchowski:
@kamsi Please, prepare test port and update JIRA tickets with those.

view/utils/init-table-sorting-on-client.js
kamsi: @mtuchowski not sure about internals, but to my understanding this assures that we will not use it by accident (which inmo is undesired).

view/utils/init-table-sorting-on-client.js
kamsi: @mtuchowski brief search did not bring anything interesting. Still it didn't work ouside of the box for dbjs Date... even though it stringified itself to that...

kamsi:
@mtuchowski it passed the tests, so I think we can merge.

mtuchowski:
@kamsi Please resolve conflicts.

PR #1893, opened 2017-06-01
WS communication utilities

kamsi:
This set of tools enables easier handling of WS communication.


kamsi:
@roxanadina seems safe to merge, right?

PR #1896, opened 2017-06-06
Add sorting feature to "Times / per person" tables

ilja903:
https://unctad.atlassian.net/browse/EV2CORE-105

### DIFF
--- a/view/utils/init-table-sorting-on-client.js
+++ b/view/utils/init-table-sorting-on-client.js
@@ -34,9 +34,22 @@ module.exports = function (domElement/*, opts */) {
 		type: 'numeric'
 	});
 	var opts = normalizeOptions(arguments[1]);
-	setInterval(function () {
-		var element = window.jQuery(domElement);
-		element.trigger('update');
-		element.tablesorter(opts);
-	}, 500);
+	var additionalOpts = normalizeOptions(arguments[2]);
+
+	if(additionalOpts && additionalOpts.beforeSortStartFn){
+		setTimeout(function () {
+			var element = window.jQuery(domElement);
+			element.bind("sortStart",function() {
+				additionalOpts.beforeSortStartFn(element)
+			});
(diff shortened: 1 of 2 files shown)


view/utils/init-table-sorting-on-client.js
```diff
@@ -34,9 +34,22 @@ module.exports = function (domElement/*, opts */) {
...
 	});
 	var opts = normalizeOptions(arguments[1]);
-	setInterval(function () {
-		var element = window.jQuery(domElement);
-		element.trigger('update');
-		element.tablesorter(opts);
-	}, 500);
+	var additionalOpts = normalizeOptions(arguments[2]);
+
+	if (additionalOpts && additionalOpts.beforeSortStartFn) {
```
kamsi: It looks like we count on our luck here. We use `setTimeout instead of `setInterval` to setup the whole tablesorter...
If I understand correctly we introduced using `setInterval` to avoid problems with `setTimeout` (as `setTimeout` could simply be not enough)...

I don't see a reason for changing the way we setup sorter for that particular case. So what was the problem here? You didn't want bind for the second time? In other words: Will this such code work:

```javascript

setInterval(function () {
var element = window.jQuery(domElement);
if (additionalOpts && additionalOpts.beforeSortStartFn) {
 element.bind("sortStart", function () {
 additionalOpts.beforeSortStartFn(element);
});
}
element.trigger('update');
element.tablesorter(opts);
}, 500);
```

ilja903: no. I made this decision because element.trigger('update'); always executes sortStart. Because of that when you open a detailrow in a table it gets sorted after 500ms... Basically we are relying on luck here. To my "defence" I can say that this example luckily worked well on http://ilja9032-els.eregistrations.org/time/per-person/ with production db

kamsi:
@mtuchowski @roxanadina I would like you to take a look as well.

PR #1897, opened 2017-06-06
Dashboard table 1 : all the files that are either in status Withdraw_0 or Withdraw_1

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-141

### DIFF
--- a/server/routes/statistics.js
+++ b/server/routes/statistics.js
@@ -204,7 +204,8 @@ module.exports = function (config) {
 					if (!resultServiceData) {
 						resultServiceData = result.byService[serviceName] = {};
 					}
-					var count = periodData.atPickupWithdrawnOrClosed || 0;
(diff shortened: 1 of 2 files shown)


server/routes/statistics.js
```diff
@@ -585,73 +602,12 @@ module.exports = function (config) {
 					result.chartsResult   = chartsResult;
 					return getFilesCompleted(query).then(function (res) {
 						result.filesCompleted = res;
+						return deferred(true);
```
mtuchowski: You could return `query` here and in next line just chain `getApprovedCerts` call like this:

```javascript
    return query;
}).then(getApprovedCerts).then(function (res) {
```

One less block level. But not a deal breaker.

kamsi:
@mtuchowski not that it was actually tested... but I will put it on the test port none the less... if sth pops up, we can either do a quick-fix, or revert...

PR #1898, opened 2017-06-08
Move Files / Certificates

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-150


routes/statistics.js
```diff
@@ -54,6 +48,7 @@ module.exports = function () {
 		},
 		rejections: require('../view/statistics-rejections'),
 		accounts: require('../view/statistics-accounts'),
+		old: require('../view/statistics-old'),
```
mtuchowski: image

PR #1901, opened 2017-06-13
Service enabling resending of requests

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-153

### DIFF
--- a/server/services/ws-services-requests-manager.js
+++ b/server/services/ws-services-requests-manager.js
@@ -33,7 +33,9 @@ exports.send = function (data) {
 
 exports.resendErrored = function () {
 	return requestLogger.getErrored().then(function (errored) {
-		return deferred.map(errored, function (requestData) {
+		return deferred.map(errored.filter(function (item) {
+			return item.requestType === 'sender';
+		}), function (requestData) {
 			return exports.send(requestData);
 		});
 	});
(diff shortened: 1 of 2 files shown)


server/services/ws-services-requests-manager.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+var requestLogger = require('../../webservice/request-logger')
+  , deferred = require('deferred')
+  , resolveProjectRoot = require('cjs-module/resolve-project-root')
+  , path        = require('path')
+  , projectRoot = process.cwd()
+  , debug       = require('debug-ext')('ws-setrvices-requests-manager');
+
+exports.getRequestsDirectory = function () {
```
roxanadina: @kamsi i think this should be memoized, we will probably use this service more often than at every restart (to be decided however), while the file structure won t change during app run
(though not super performance booster :))

server/services/ws-services-requests-manager.js
```diff
@@ -0,0 +1,42 @@
...
+		if (!data.requestHandler) {
+			debug('Broken logger data, attempting send with no requestHandler specified');
+			return;
+		}
+		requestHandler = require(path.resolve(dir, data.requestHandler));
+		if (!requestHandler) {
+			debug('No request handler defined for %s, aborting send.', data.requestHandler);
+			return;
+		}
+		return requestHandler(data);
```
roxanadina: also check it s a function?

PR #1903, opened 2017-06-26
Expose keyPath on fields

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-169


kamsi:
@mtuchowski the name is needed in the result of `toWebServiceJSON` after all: https://github.com/egovernment/eregistrations/pull/1903/files#diff-da236ed38803988d1fb086a38b5db94cR242

PR #1916, opened 2017-07-19
Rearrange the dashboard

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-164


mtuchowski:
Prepare test port with probably ELS and update the ticket.

PR #1917, opened 2017-07-20
Sent back for correction table

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-165

PR #1918, opened 2017-07-21
Processing times per role overview

kamsi:
https://unctad.atlassian.net/browse/EV2CORE-167


kamsi:
@mtuchowski assumes that https://github.com/egovernment/eregistrations/pull/1917 is merged

kamsi:
@mtuchowski already exposed here: http://kamsi1-ldz.eregistrations.org/, so you can just pass it to tests.

PR #1920, opened 2017-07-28
Enable direct filter resolution in official access

kamsi:


kamsi:
Needed for: https://github.com/egovernment/eregistrations-salvador/pull/1961

kamsi:
Also needed for: https://github.com/egovernment/eregistrations-salvador/pull/1960

PR #1930, opened 2017-09-21
Shorten timeouts for ws on attempt to get wsdl

kamsi:
Solves: https://unctad.atlassian.net/browse/ELS-242


kamsi:
So all that is needed in an end system now, is a an entry in the env of the sort:

```javascript
soapClientOptions: {
		wsdl_options: {
			timeout: 10000
		}
	}
```
and this particular one will ensure that we will not try to get initial wsdl for more than 10 seconds

roxanadina:
@kamsi lint

kamsi:
@roxanadina  it was a failing test, I believe I fixed that, though I don't have time to wait for Circle CI now...

PR #1941, opened 2017-11-20
OAuth integration

mtuchowski:
* https://unctad.atlassian.net/browse/EV2CORE-197 - Disable login if oauth is present and register routes
* https://unctad.atlassian.net/browse/EV2CORE-198 - Include oAuth auth
* https://unctad.atlassian.net/browse/ELS-294 - User data needed for v2 to still exist: handle creating new user in v2 upon creating users in v4. get email from the token : check if it is existent in v2. If not, request more user data from v4 and create it in v2.
* https://unctad.atlassian.net/browse/ELS-291 - Users admin new users are treated as normal usersdisable v2 profile data - you have to modify it in v4
* https://unctad.atlassian.net/browse/ELS-292 - My profile: redirect to CAS

### DIFF
--- a/server/oauth.js
+++ b/server/oauth.js
@@ -6,34 +6,18 @@ var debug          = require('debug-ext')('oauth')
   , assign         = require('es5-ext/object/assign')
   , userEmailMap   = require('mano/lib/server/user-email-map')
   , authentication = require('mano-auth/server/authentication')
-  , loadToMemoryDb = require('mano/lib/server/resolve-user-access')
   , serializeValue = require('dbjs/_setup/serialize/value')
   , generateUnique = require('time-uuid')
   , request        = require('request')
   , jwtDecode      = require('jwt-decode')
-  , env            = mano.env
-  , userStorage    = mano.dbDriver.getStorage('user');
-
-var dbjsDataRecord = function (id, value) {
-	return { id: id, data: { value: serializeValue(value) } };
-};
+  , env            = mano.env;
 
 var createUser = function (data) {
 	return mano.queryMemoryDb([], 'addUser', JSON.stringify(data));
 };
 
-var updateUser = function (userId, data) {
-	var records = [];
-
-	Object.keys(data).forEach(function (key) {
-		records.push(dbjsDataRecord(userId + '/' + key, data[key]));
-	});
-
-	return userStorage.storeMany(records)(function () {
-		return loadToMemoryDb([userId]);
-	})(function () {
-		return userId;
-	});
+var registerDemoUser = function (userId, data) {
+	return mano.queryMemoryDb([userId], 'registerDemoUser', { userId: userId, data: data });
 };
 
 var generateUrl = function (path, query) {
@@ -308,8 +292,7 @@ module.exports = exports = {
 						});
 					}
 
-					return updateUser(demoUserId, {
-						isDemo: undefined,
+					return registerDemoUser(demoUserId, {
 						firstName: decoded.fname,
 						lastName: decoded.lname,
 						email: decoded.email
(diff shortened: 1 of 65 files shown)


server/oauth.js
```diff
@@ -0,0 +1,170 @@
...
+					});
+				}).done(function (userId) {
+					if (!decoded.email_verified) {
+						res.writeHead(302, { Location: '/request-confirm-account/' });
+						res.end();
+						return;
+					}
+
+					login(userId, req, res);
+
```
kamsi: Let's pretty please ensure we have some handle on the token to issue further requests, like: https://github.com/egovernment/eregistrations-salvador/pull/2071/files#diff-fb19faa00022c30c2cb25d27383803fcR121

view/users-table.js
```diff
@@ -53,8 +54,10 @@ exports._actionsColumn = {
 	head: th({ class: 'submitted-user-data-table-actions' }),
 	data: function (user) {
 		var isSelfUser = (user === this.user);
+
 		return td({ class: 'actions' },
-			a({ href: isSelfUser ? '/profile/' : url('user', user.__id__) },
+			a({ href: isSelfUser ? (externalAuthentication.profilePage || '/profile/')
```
kamsi: I propose adding ```target: "_blank"``` attribute, but this is merely a proposal

server/oauth.js
```diff
@@ -0,0 +1,299 @@
...
+	});
+
+	return userStorage.storeMany(records)(function () {
+		return loadToMemoryDb([userId]);
+	})(function () {
+		return userId;
+	});
+};
+
+var generateUrl = function (path, query) {
```
kamsi: I guess this could be general purpose util after all :)

server/oauth.js
```diff
@@ -0,0 +1,299 @@
...
+				method: 'GET',
+				headers: {
+					Authorization: 'Bearer ' + accessToken
+				}
+			}, function (error, response, body) {
+				if (error) {
+					debug('Error received from invalidation endpoint:', error);
+				}
+			});
+		}
```
kamsi: I believe it would be safer to put an `else` here (or early return before this point)... in that `else` we would do `next` and the other `next` would be in the body of the callback from request... otherwise it is quite easy to miss sth and make a mistake by mutating sth in the callback and forgetting about handling `next` properly.

server/oauth.js
```diff
@@ -0,0 +1,299 @@
...
+						});
+					}
+
+					return updateUser(demoUserId, {
+						isDemo: undefined,
+						firstName: decoded.fname,
+						lastName: decoded.lname,
+						email: decoded.email
+					});
+				}).done(function (userId) {
```
kamsi: Hmmm... so here we use promise inside a callback... from what I remember this is not the best practice. Maybe you could wrap request in a promise and chain the next promise with it?

server/oauth.js
```diff
@@ -0,0 +1,299 @@
...
+					} else {
+						debug('Failed to refresh token:', response.body);
+						// debug('Failed to refresh token:', response.statusCode);
+					}
+				});
+			}
+		}
+
+		// 2. Passthru to other middlewares.
+		next();
```
kamsi: As discussed on skype this needs to be changed.

server/oauth.js
mtuchowski: As request module does not conform to standard node way of doing callbacks (in respect to input arguments) this is proper (and working) way to do it. Nothing especially cringy about it (maybe to purists ;) ).

server/oauth.js
mtuchowski: Propose where do you want to keep it.

PR #1942, opened 2017-11-24
Allow manager to change his v2 profile

kamsi:
https://unctad.atlassian.net/browse/ELS-308

### DIFF
--- a/server/oauth.js
+++ b/server/oauth.js
@@ -294,27 +294,29 @@ module.exports = exports = {
 				userEmailMap(function (map) {
 					return map.get(serializeValue(decoded.email));
 				})(function (userId) {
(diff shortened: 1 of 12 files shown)


server/oauth.js
```diff
@@ -249,25 +249,27 @@ module.exports = exports = {
 				userEmailMap(function (map) {
 					return map.get(serializeValue(decoded.email));
 				})(function (userId) {
-					if (userId) return userId;
-
-					var isPublicApp = req.$appName === 'public'
-					  , demoUserId  = isPublicApp ? null : res.cookies.get('demoUser');
```
mtuchowski: Demo user handling is still needed here.

